PR #36, opened 2019-10-20
Migrate to go modules

nickborysov:


admiral/cmd/admiral/cmd/root.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/admiral/admiral/pkg/clusters"
+	"github.com/istio-ecosystem/admiral/admiral/pkg/clusters"
```
aattuluri: @nickborysov Is there a specific reason why `istio-ecosystem`  (git org name) was added to the package structure?

nickborysov: @aattuluri this is usual for go modules. When you call `go mod init`, it will create module like github.com/<account_name>/<repo_name>.

PR #42, opened 2019-12-06
Serviceentry to use a configmap for unique address cache

josephpeacock:
We noticed a bug where serviceentries are occasionally created with duplicate IPs. This is fixing that.

Still pending validation + writing tests.


aattuluri:
lgtm

PR #45, opened 2019-12-26
Making labels configurable

josephpeacock:
Also added a target group label to the setup that can eventually be used to route traffic between deployments based on an arbitrary label 

Signed-off-by: Joe Peacock <joe_peacock@intuit.com>

### DIFF
--- a/admiral/cmd/admiral/cmd/root.go
+++ b/admiral/cmd/admiral/cmd/root.go
@@ -71,11 +71,11 @@ func GetRootCmd(args []string) *cobra.Command {
 		"Prefix to use when creating SAN for Service Entries")
 	rootCmd.PersistentFlags().StringVar(&params.SecretResolver, "secret_resolver", "",
 		"Type of resolver to use to fetch kubeconfig for monitored clusters")
-	rootCmd.PersistentFlags().StringVar(&params.LabelSet.DeploymentLabel, "deployment_label", "istio-injected",
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.DeploymentLabel, "deployment_label", "sidecar.istio.io/inject",
 		"The label, on a deployment, which must be set to \"True\" for Admiral to listen on the deployment")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.SubsetLabel, "subset_label", "subset",
 		"The label, on a deployment, tells admiral which target group this deployment is a part of. Used for traffic splits via the Global Traffic Policy object")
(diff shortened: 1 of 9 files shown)


admiral/cmd/admiral/cmd/root.go
```diff
@@ -71,6 +71,14 @@ func GetRootCmd(args []string) *cobra.Command {
 		"Prefix to use when creating SAN for Service Entries")
 	rootCmd.PersistentFlags().StringVar(&params.SecretResolver, "secret_resolver", "",
 		"Type of resolver to use to fetch kubeconfig for monitored clusters")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.DeploymentLabel, "deployment_label", "istio-injected",
+		"The label, on a deployment, which must be set to \"True\" for Admiral to listen on the deployment")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.TargetGroupLabel, "target_group_label", "target-group",
```
aattuluri: As we spoke, change this to `subset_label`

aattuluri: As we discussed add another label for admiral to ignore a deployment.

admiral/cmd/admiral/cmd/root.go
```diff
@@ -71,6 +71,14 @@ func GetRootCmd(args []string) *cobra.Command {
...
+		"The label, on a deployment, tells admiral which target group this deployment is a part of. Used for traffic splits via the Global Traffic Policy object")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.TargetGroupLabel, "namespace_injected_label", "sidecar.istio.io/inject",
+		"The label key, on a namespace, which tells Istio to perform sidecar injection")
```
aattuluri: `sidecar.istio.io/inject` is pod annotation, not namespace?

aattuluri: Probably this should be `istio-injection`: `enabled`

admiral/cmd/admiral/cmd/root.go
```diff
@@ -71,6 +71,14 @@ func GetRootCmd(args []string) *cobra.Command {
 		"Prefix to use when creating SAN for Service Entries")
 	rootCmd.PersistentFlags().StringVar(&params.SecretResolver, "secret_resolver", "",
 		"Type of resolver to use to fetch kubeconfig for monitored clusters")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.DeploymentLabel, "deployment_label", "istio-injected",
```
aattuluri: This should be `sidecar.istio.io/inject: true` annotation (which will also fix the bug)?

PR #48, opened 2019-12-31
Adding label to skip admiral generated istio resource update

vrushalijoshi:
Added "disable-update" label to skip any updates to istio resource

### DIFF
--- a/admiral/pkg/clusters/registry.go
+++ b/admiral/pkg/clusters/registry.go
@@ -1026,18 +1026,28 @@ func addUpdateIstioResource(rc *RemoteController, m istio.Config, exist *istio.C
 		_, e = rc.IstioConfigStore.Create(m)
 
 	} else {
-		exist.Labels = m.Labels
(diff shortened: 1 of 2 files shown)


admiral/pkg/clusters/registry.go
```diff
@@ -1026,18 +1026,28 @@ func addUpdateIstioResource(rc *RemoteController, m istio.Config, exist *istio.C
 		_, e = rc.IstioConfigStore.Create(m)
 
 	} else {
-		exist.Labels = m.Labels
-		exist.Spec = m.Spec
-		exist.Annotations = m.Annotations
-		verb = "Update"
-		_, e = rc.IstioConfigStore.Update(*exist)
+		if len(exist.Labels) > 0 {
+			if _, ok := exist.Labels["disable-update"]; ok {
```
virajrk: Can you please change something to indicate for admiral? Because `disable-update` is generic label. Like `disable-admiral-update` WDUT?

PR #55, opened 2020-01-13
Added secret update function

gaopan233:
Added secret update function
Modified logic when secret exists: secrets will always be reloaded


admiral/pkg/controller/secret/secretcontroller.go
```diff
@@ -139,6 +145,13 @@ func NewController(
 				queue.Add(key)
 			}
 		},
+		UpdateFunc: func(oldObj interface{}, newObj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(newObj)
+			log.Infof("Processing update: %s", key)
```
aattuluri, line 150: Change the log line to "Processing secret update for cluster: %s" for better readability.

PR #59, opened 2020-01-21
Fix local SE generation. Fix related bugs.

aattuluri:


admiral/pkg/controller/common/common.go
```diff
@@ -66,6 +64,18 @@ func GetCname(deployment *k8sAppsV1.Deployment, identifier string, nameSuffix st
 	return environment + Sep + alias + Sep + nameSuffix
 }
 
+func GetEnv(deployment *k8sAppsV1.Deployment) string {
+	var environment = deployment.Spec.Template.Labels[Env]
```
josephpeacock: Do we want to also check annotations before falling back on ns splitting?

aattuluri: Done. Good suggestion, keeping it consistent with `identifier`. At some point we will move to selector which will be based of labels, so we should still encourage labels over annotations.

PR #68, opened 2020-02-12
Formalizing the behavior to fall back to annotation if label isn't pr…

josephpeacock:
…esent

Signed-off-by: Joe Peacock <joe_peacock@intuit.com>


admiral/pkg/controller/common/common.go
```diff
@@ -98,3 +95,12 @@ func GetNodeLocality(node *k8sV1.Node) string {
 	region, _ := node.Labels[NodeRegionLabel]
 	return region
 }
+
+func GetValueForKeyFromDeployment(key string, deployment *k8sAppsV1.Deployment) string {
```
aattuluri: Can you check if we need to use this in deployment.go?

josephpeacock: This specifically isn't, but the 'identity' label was hardcoded in a number of places. Removed those.

admiral/pkg/controller/common/identity.go
```diff
@@ -0,0 +1,16 @@
+package common
+
+import "istio.io/istio/pkg/log"
+
+var workloadIdentifier = Identity
```
aattuluri: I am assuming the intention here is to manage the overriden Identity instead of passing around AdmiralParams? If so, can we have params here instead of just `workloadIdentifier`. That way we can consume any overriden parameters via this file/param.

PR #84, opened 2020-03-23
Update sidecar resource in workload namespace

vrushalijoshi:
Code changes to update the sidecar resource with the workload namespace

### DIFF
--- a/admiral/cmd/admiral/cmd/root.go
+++ b/admiral/cmd/admiral/cmd/root.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	ctx, cancel    = context.WithCancel(context.Background())
+	ctx, cancel = context.WithCancel(context.Background())
 )
 
 // GetRootCmd returns the root of the cobra command-tree.
@@ -85,6 +85,8 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The workload identity  key, on deployment which holds identity value used to generate cname by admiral. Default label key will be \"identity\" Admiral will look for a label with this key. If present, that will be used. If not, it will try an annotation (for use cases where an identity is longer than 63 chars)")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GlobalTrafficDeploymentLabel, "globaltraffic_deployment_label", "identity",
 		"The label key which will be used to tie globaltrafficpolicy objects to deployments. Configured separately to the workload identity key because this one won't fall back to annotations.")
+	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarUpdates, "workload_sidecar_updates", "disabled",
+		"The parameter will be used to decide whether to update workload sidecar resource or not. By default these updates will be disabled.")
 
 	return rootCmd
 }
--- a/admiral/cmd/admiral/cmd/root.go
+++ b/admiral/cmd/admiral/cmd/root.go
@@ -85,8 +85,10 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The workload identity  key, on deployment which holds identity value used to generate cname by admiral. Default label key will be \"identity\" Admiral will look for a label with this key. If present, that will be used. If not, it will try an annotation (for use cases where an identity is longer than 63 chars)")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GlobalTrafficDeploymentLabel, "globaltraffic_deployment_label", "identity",
 		"The label key which will be used to tie globaltrafficpolicy objects to deployments. Configured separately to the workload identity key because this one won't fall back to annotations.")
-	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarUpdates, "workload_sidecar_updates", "disabled",
+	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarUpdate, "workload_sidecar_update", "disabled",
 		"The parameter will be used to decide whether to update workload sidecar resource or not. By default these updates will be disabled.")
+	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarName, "workload_sidecar_name", "default",
+		"Name of the sidecar resource in the workload namespace. By default sidecar resource will be named as \"default\".")
 
 	return rootCmd
 }
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -201,7 +201,7 @@ func modifySidecarForLocalClusterCommunication(sidecarNamespace string, dependen
 
 	sidecar, _ := sidecarConfig.IstioClient.NetworkingV1alpha3().Sidecars(sidecarNamespace).Get(common.GetWorkloadSidecarName(), v12.GetOptions{})
 
-	if sidecar == nil {
+	if sidecar == nil || sidecar.Spec.Egress == nil {
 		return
 	}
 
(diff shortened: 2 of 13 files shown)


admiral/cmd/admiral/cmd/root.go
```diff
@@ -85,6 +85,8 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The workload identity  key, on deployment which holds identity value used to generate cname by admiral. Default label key will be \"identity\" Admiral will look for a label with this key. If present, that will be used. If not, it will try an annotation (for use cases where an identity is longer than 63 chars)")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GlobalTrafficDeploymentLabel, "globaltraffic_deployment_label", "identity",
 		"The label key which will be used to tie globaltrafficpolicy objects to deployments. Configured separately to the workload identity key because this one won't fall back to annotations.")
+	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarUpdates, "workload_sidecar_updates", "disabled",
```
aattuluri: Call this `workload_sidecar_update` with out the `s`.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -189,19 +190,98 @@ func createServiceEntryForNewServiceOrPod(env string, sourceIdentity string, rem
 			//} else {
 			//	log.Errorf(LogErrFormat, "Create", istioModel.VirtualService.Type, virtualServiceName, rc.ClusterID, err)
 			//}
+			if common.GetWorkloadSidecarUpdates() == "enabled" {
+				for _, val := range dependents.Map() {
```
aattuluri: I think this `for` loop should be outside the if block since the DependencyNamespaceCache can be used for other purposes.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -189,19 +190,98 @@ func createServiceEntryForNewServiceOrPod(env string, sourceIdentity string, rem
...
 
+func modifySidecarForLocalClusterCommunication(sidecarNamespace string, dependencyNamespace string, localFqdn string, rc *RemoteController) {
+
+	//get existing sidecar from the cluster
+	sidecarConfig := rc.SidecarController
+
+	if sidecarConfig == nil {
+		return
+	}
+	// sidecarConfig := rc.IstioConfigStore.Get(istioModel.Sidecar.Type, SidecarName, sidecarNamespace)
```
aattuluri: Remove the commented out code.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -189,19 +190,98 @@ func createServiceEntryForNewServiceOrPod(env string, sourceIdentity string, rem
...
+
+	//sidecar := &networking.Sidecar{}
+
+	sidecar, _ := sidecarConfig.IstioClient.NetworkingV1alpha3().Sidecars(sidecarNamespace).Get(SidecarName, v12.GetOptions{})
```
aattuluri: If sidecar is `nil` it should return from the method. We should only modify a sidecar if one exists. Cannot really create one and start restricting since that might break other functionality on the workload.

admiral/cmd/admiral/cmd/root.go
aattuluri: Add another parameter `workload_sidecar_name` to configure the sidecar name. The value should be `default` by default.

josephpeacock: Can we be more explicit about what the `disabled` state is? Would that just mean every sidecar has `*/*`? (And is there any configuration that needs to be done to make it be that way?)

vrushalijoshi: sure, will update this

admiral/pkg/clusters/serviceentry.go
```diff
@@ -175,33 +176,82 @@ func createServiceEntryForNewServiceOrPod(env string, sourceIdentity string, rem
...
+	//get existing sidecar from the cluster
+	sidecarConfig := rc.SidecarController
+
+	if sidecarConfig == nil {
+		return
+	}
+
+	sidecar, _ := sidecarConfig.IstioClient.NetworkingV1alpha3().Sidecars(sidecarNamespace).Get(common.GetWorkloadSidecarName(), v12.GetOptions{})
+
+	if sidecar == nil || sidecar.Spec.Egress == nil {
```
aattuluri: Make sure `sidecar.Spec` always returns a value when `sidecar` is not nil.

vrushalijoshi: Yes, verified it as part of test - TestModifyNonExistingSidecarForLocalClusterCommunication

Spec is present when sidecar is not nil

admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -268,17 +278,78 @@ import (
...
+
+	sidecarController, _ := istio.NewSidecarController(stop, &handler, config, time.Duration(1000))
+
+	remoteController := &RemoteController{SidecarController: sidecarController}
+
+	modifySidecarForLocalClusterCommunication("test-sidecar-namespace", "test-dependency-namespace", "test-local-fqdn", remoteController)
+
+	sidecarObj, _ := sidecarController.IstioClient.NetworkingV1alpha3().Sidecars("test-sidecar-namespace").Get(common.GetWorkloadSidecarName(), v12.GetOptions{})
+
+	if sidecarObj == nil || sidecarObj.Spec.Egress != nil {
```
josephpeacock: Can you be a little more specific with the assertion here? What should `sidecarObj.Spec.Egress` be?

admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -268,17 +278,78 @@ import (
...
+		updatedSidecar, error := sidecarController.IstioClient.NetworkingV1alpha3().Sidecars("test-sidecar-namespace").Get("default", v12.GetOptions{})
+
+		if error != nil || updatedSidecar == nil {
+			t.Fail()
+		}
+
+		if len(updatedSidecar.Spec.Egress[0].Hosts) != 2 {
+			t.Fail()
+		}
+	}
```
josephpeacock: else fail the test with an error?

admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -268,17 +278,78 @@ import (
...
+
+		modifySidecarForLocalClusterCommunication("test-sidecar-namespace", "test-dependency-namespace", "test-local-fqdn", remoteController)
+
+		updatedSidecar, error := sidecarController.IstioClient.NetworkingV1alpha3().Sidecars("test-sidecar-namespace").Get("default", v12.GetOptions{})
+
+		if error != nil || updatedSidecar == nil {
+			t.Fail()
+		}
+
+		if len(updatedSidecar.Spec.Egress[0].Hosts) != 2 {
```
josephpeacock: Same as above - in my opinion the tests are the most valuable when we're asserting for exactly what's needed. If there's some sidecar egress formatting bug introduced later, we want these tests to catch it

admiral/pkg/clusters/serviceentry_test.go
vrushalijoshi: This will change based on what is configured in the workload namespace. For this test, assumption is that no sidecar resource is configured. So test is that we are not creating this resource.

admiral/pkg/clusters/serviceentry_test.go
vrushalijoshi: Agreed, modified the test

PR #88, opened 2020-03-27
Handle dependency update event

aattuluri:


admiral/pkg/controller/common/types.go
```diff
@@ -22,7 +22,7 @@ type SidecarEgress struct {
 }
 
 type SidecarEgressMap struct {
-	cache map[string][]SidecarEgress
```
josephpeacock: It looks to me like this is now mapping the identity value -> a map of namespace -> sidecar egress object. Can you add a comment to make that clear?

admiral/pkg/controller/common/types.go
```diff
@@ -137,13 +137,13 @@ func (s *SidecarEgressMap) Put(asset string, namespace string, fqdn string) {
 	s.mutex.Lock()
 	var mapVal = s.cache[asset]
 	if mapVal == nil {
-		mapVal = make([]SidecarEgress, 0)
```
josephpeacock: Can we change the variable name from `asset` to `identity`?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -208,10 +206,11 @@ func modifySidecarForLocalClusterCommunication(sidecarNamespace string, dependen
 	//copy and add our new local FQDN
 	newSidecar := copySidecar(sidecar)
 
-	egressHost := dependencyNamespace + "/" + localFqdn
-
-	if !util.Contains(newSidecar.Spec.Egress[0].Hosts, egressHost) {
-		newSidecar.Spec.Egress[0].Hosts = append(newSidecar.Spec.Egress[0].Hosts, egressHost)
+	for _, sidecarEgress := range sidecarEgressMap {
```
josephpeacock: I'm not against changing the SidecarEgressMap to be a map of maps, but you're not leveraging that here at all. You're still doing get(identity) then iterating through the values.

I guess my question then is why make the change to make the value another map, rather than a slice?

aattuluri: `State before`
EgressMap had a mapping from a service identity -> List of Egresses
Every time we process a service, we add an entry into the List of Egress for all its dependents, meaning the list just keeps growing with duplicates
Now the modifySIdecar, the way it was being called also had a problem, you would call it per egress entry for the same sidecar N number of times, where N is size of the egress list

`State after this change`
EgressMap had a mapping from a service identity -> Map of Egresses by namespace
Every time we process a service, we add an entry and duplicates will be filtered
modifySidecar still has to iterate over all egress entries but only the ones not already present will be added and sidecar will be updated only once per source service

admiral/pkg/controller/common/types.go
aattuluri: Done

admiral/pkg/controller/common/types.go
aattuluri: Done

PR #91, opened 2020-04-07
Traffic at ingressgateway should always be routed to cluster local service instance

aattuluri:


admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -469,3 +469,72 @@ func TestCreateServiceEntry(t *testing.T) {
 	}
 
 }
+
+func TestCreateIngressOnlyVirtualService(t *testing.T) {
+
```
josephpeacock, line 474: This test could be cleaned up a bit using the iterative test format: https://gist.github.com/josephpeacock/6f5a73c3f133029d59ab736c83c1b087 - let's you avoid duplicating the validation logic and makes it easier to add cases in the future if needed

admiral/pkg/controller/admiral/dependency_test.go
```diff
@@ -20,3 +23,76 @@ func TestNewDependencyController(t *testing.T) {
 		t.Errorf("Dependency controller should never be nil without an error thrown")
 	}
 }
+
+func TestGetDependencies(t *testing.T) {
+	stop := make(chan struct{})
+	handler := test.MockDependencyHandler{}
+
+	dependencyController, err := NewDependencyController(stop, &handler, "../../test/resources/admins@fake-cluster.k8s.local", "ns", time.Duration(1000))
```
josephpeacock, line 31: You can use `Common.GetKubeconfigPath()` here - we already have a fake config loaded in for testing

admiral/pkg/controller/admiral/dependency_test.go
```diff
@@ -20,3 +23,76 @@ func TestNewDependencyController(t *testing.T) {
...
+
+	if err != nil {
+		t.Errorf("Unexpected err %v", err)
+	}
+
+	if dependencyController == nil {
+		t.Errorf("Dependency controller should never be nil without an error thrown")
+	}
+
+	deps, _  := dependencyController.GetDependencies()
```
josephpeacock, line 41: Can you add a case to this test where there are dependencies to get? Otherwise we're not really testing the functionality, as 

```
func (d *DependencyController) GetDependencies() ([]v1.Dependency, error) {
  return nil, nil
} 
```
also passes this test.

admiral/pkg/controller/admiral/dependency_test.go
```diff
@@ -20,3 +23,76 @@ func TestNewDependencyController(t *testing.T) {
...
+		t.Errorf("dep add failed")
+	}
+
+	//test update
+	updatedDep := model.Dependency{IdentityLabel: "identity", Destinations:[]string{"greeting", "payments", "newservice"}, Source: "webapp"}
+	dependencyController.Updated(makeK8sDependencyObj(depName, "namespace1", updatedDep), depObj)
+
+	updatedDepObj := dependencyController.Cache.Get(depName)
+
+	if updatedDepObj == nil || updatedDepObj.Spec.Source != "webapp" ||  len(updatedDepObj.Spec.Destinations) != 3 {
```
josephpeacock, line 81: You can use the `cmp` library for doing deep equals of objects - it could be useful here instead of comparing the different fields on their own

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +28,88 @@ func TestNewGlobalTrafficController(t *testing.T) {
 		t.Errorf("GlobalTraffic controller should never be nil without an error thrown")
 	}
 }
+
+func TestGlobalTrafficAddUpdateDelete(t *testing.T) {
+	config, err := clientcmd.BuildConfigFromFlags("", "../../test/resources/admins@fake-cluster.k8s.local")
```
josephpeacock, line 33: Same as above - can use Common.GetKubeconfigPath()

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +28,88 @@ func TestNewGlobalTrafficController(t *testing.T) {
 		t.Errorf("GlobalTraffic controller should never be nil without an error thrown")
 	}
 }
+
+func TestGlobalTrafficAddUpdateDelete(t *testing.T) {
```
josephpeacock, line 32: This test isn't asserting anything. Is there a way we can build out a GlobalTrafficHandler that actually has to handle the update? Something like what we have here? https://github.com/istio-ecosystem/admiral/blob/37deea2910e06ea46a32bef2ac308c6c834efd8b/admiral/pkg/clusters/types_test.go#L45

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +28,88 @@ func TestNewGlobalTrafficController(t *testing.T) {
...
+
+	if err != nil {
+		t.Errorf("Unexpected err %v", err)
+	}
+
+	if globalTrafficController == nil {
+		t.Errorf("GlobalTraffic controller should never be nil without an error thrown")
+	}
+
+	gtps, _  := globalTrafficController.GetGlobalTrafficPolicies()
```
josephpeacock, line 79: Same comment as `TestGetDependencies`

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +28,88 @@ func TestNewGlobalTrafficController(t *testing.T) {
...
+
+	if err != nil {
+		t.Errorf("Unexpected err %v", err)
+	}
+
+	if globalTrafficController == nil {
+		t.Errorf("GlobalTraffic controller should never be nil without an error thrown")
+	}
+
+	gtps  := globalTrafficController.GetGTPByLabel("payments", "namespace1")
```
josephpeacock, line 104: TestGetDependencies

install/admiral/overlays/demosinglecluster/kustomization.yaml
```diff
@@ -1,17 +1,15 @@
-apiversion: kustomize.config.k8s.io/v1beta1
+apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 images:
-  - name: docker.io/admiralproj/admiral
-    newTag: latest
+- name: docker.io/admiralproj/admiral
+  newName: docker.io/admiralproj/admiral
+  newTag: kubecon
```
josephpeacock, line 7: Do we want to use the kubecon tag?

admiral/pkg/controller/istio/serviceentry_test.go
```diff
@@ -24,4 +27,12 @@ func TestNewServiceEntryController(t *testing.T) {
 	if serviceEntryController == nil {
 		t.Errorf("ServiceEntry controller should never be nil without an error thrown")
 	}
+
+	serviceEntry := &v1alpha32.ServiceEntry{Spec: v1alpha3.ServiceEntry{}, ObjectMeta: v1.ObjectMeta{Name: "se1", Namespace: "namespace1"}}
```
josephpeacock, line 31: If we're going to test added/updated/deleted, can we use the real SE handler instead of the mock one? (Same comment applies for sidecars/VS as well)

admiral/pkg/controller/admiral/globaltraffic_test.go
aattuluri: I would keep the tests for controller in here, and the tests for the handler in the handler tests. If the controller isn't doing anything other than passing the object to the handler, then the only thing to be tested here is to make sure the handler is being called with the object.

admiral/pkg/controller/admiral/dependency_test.go
aattuluri: This doesn't work unless every package has `testdata/fakeconfig` as the file path is relative.

admiral/pkg/clusters/serviceentry_test.go
aattuluri: Good suggestion. Updated.

admiral/pkg/controller/admiral/dependency_test.go
aattuluri: Done.

admiral/pkg/controller/istio/serviceentry_test.go
aattuluri: Same as the comment above, this is to test the controller and not the handler. Those tests need to be added in the handler.

install/admiral/overlays/demosinglecluster/kustomization.yaml
aattuluri: Good catch, changing it back to latest.

admiral/pkg/controller/admiral/dependency_test.go
```diff
@@ -20,3 +24,88 @@ func TestNewDependencyController(t *testing.T) {
...
+	}
+
+	dep := model.Dependency{IdentityLabel: "identity", Destinations:[]string{"greeting", "payments", "newservice"}, Source: "webapp"}
+	depObj := makeK8sDependencyObj("mydep", "namespace1", dep)
+
+	dependencyController.DepCrdClient.AdmiralV1().Dependencies("namespace1").Create(depObj);
+
+	deps, _  = dependencyController.GetDependencies()
+
+	if deps == nil || len(deps) == 0{
```
josephpeacock, line 55: Can we assert that the objects match, rather than just non-empty?

admiral/pkg/controller/admiral/globaltraffic_test.go
josephpeacock: That makes sense - my next question: can we assert that? Maybe add a field to the mock handler struct that records params? Because as it stands now, if globalTrafficController.Updated was completely empty, this would pass.

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +29,118 @@ func TestNewGlobalTrafficController(t *testing.T) {
...
+
+	globalTrafficController.CrdClient.AdmiralV1().GlobalTrafficPolicies("namespace1").Create(gtpObj);
+
+	gtps, err  = globalTrafficController.GetGlobalTrafficPolicies()
+
+	if err != nil {
+		t.Errorf("get gtps shouldn't error out, got: %v", err)
+	}
+
+	if gtps == nil || len(gtps) == 0 {
```
josephpeacock, line 100: Same as before, can we assert that the objects match?

admiral/pkg/controller/admiral/globaltraffic_test.go
```diff
@@ -25,3 +29,118 @@ func TestNewGlobalTrafficController(t *testing.T) {
...
+
+	gtpName := "gtp1"
+	gtp := model.GlobalTrafficPolicy{Selector: map[string]string{"identity": "payments", "env": "e2e"}, Policy:[]*model.TrafficPolicy{}}
+	gtpObj := makeK8sGtpObj(gtpName, "namespace1", gtp)
+
+	globalTrafficController.CrdClient.AdmiralV1().GlobalTrafficPolicies("namespace1").Create(gtpObj);
+
+	gtps  = globalTrafficController.GetGTPByLabel("payments", "namespace1")
+
+	if gtps == nil || len(gtps) == 0 {
```
josephpeacock, line 137: Can we assert the objects match?

admiral/pkg/controller/admiral/dependency_test.go
aattuluri: Done

admiral/pkg/controller/admiral/globaltraffic_test.go
aattuluri: Added a simple object tracker in the mock handler.

admiral/pkg/controller/admiral/globaltraffic_test.go
aattuluri: The fake client doesn't seem to work, will fix it as part of the CRD group migration. Issue filed.

admiral/pkg/controller/admiral/globaltraffic_test.go
aattuluri: The fake client doesn't seem to work, will fix it as part of the CRD group migration. Issue filed.

admiral/pkg/controller/admiral/dependency_test.go
aattuluri: Done

install/admiral/overlays/demosinglecluster/kustomization.yaml
aattuluri: Done

PR #107, opened 2020-05-19
Admiral now respects the ignore label on services

josephpeacock:


admiral/pkg/controller/admiral/service.go
```diff
@@ -41,6 +42,11 @@ type serviceCache struct {
 }
 
 func (s *serviceCache) Put(service *k8sV1.Service) {
+	if service.Labels[common.GetLabelSet().AdmiralIgnoreLabel] == "true" {
```
aattuluri, line 45: Can reuse the approach on deployments, https://github.com/istio-ecosystem/admiral/blob/70af3fb2459102bc31e47ecb1307b2e342435dab/admiral/pkg/controller/admiral/deployment.go#L185

Given we are not using labels for client side filters, it would be best to keep stuff in annotations and consistent

admiral/pkg/clusters/types_test.go
```diff
@@ -39,6 +39,7 @@ func init() {
 
 	p.LabelSet.WorkloadIdentityKey="identity"
 	p.LabelSet.GlobalTrafficDeploymentLabel="identity"
+	p.LabelSet.AdmiralIgnoreLabel="admiral-ignore"
```
aattuluri, line 42: Not needed any more?

admiral/pkg/controller/common/common_test.go
```diff
@@ -33,6 +33,7 @@ func init() {
 
 	p.LabelSet.WorkloadIdentityKey="identity"
 	p.LabelSet.GlobalTrafficDeploymentLabel="identity"
+	p.LabelSet.AdmiralIgnoreLabel="admiral-ignore"
```
aattuluri, line 36: Not needed any more?

admiral/pkg/controller/common/rolloutcommon_test.go
```diff
@@ -30,6 +30,7 @@ func init() {
 
 	p.LabelSet.WorkloadIdentityKey="identity"
 	p.LabelSet.GlobalTrafficDeploymentLabel="identity"
+	p.LabelSet.AdmiralIgnoreLabel="admiral-ignore"
```
aattuluri, line 33: Not needed any more?

PR #142, opened 2020-09-09
Ci add integration

Mengying-Li:


.circleci/config.yml
```diff
@@ -40,6 +42,69 @@ jobs:
...
+            sudo chown -R $(whoami) /usr/local/bin
+            curl -L -o /usr/local/bin/kustomize https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64
+            chmod +x /usr/local/bin/kustomize
+            kustomize version
+      - run:
+          name: Run Integration Test
+          command:  |
+            make gen-yaml
+            cd tests
+            ./run.sh "1.16.8" "1.5.6" "../out"
```
aattuluri, line 103: You can remove 1.5.6, 1.5.7 should be good enough

tests/run.sh
```diff
@@ -6,8 +6,16 @@ k8s_version=$1
 istio_version=$2
 install_dir=$3
 
-source ./create_cluster.sh $k8s_version "virtualbox"
-./install_istio.sh $istio_version "osx"
+# Uncomment below to run integration test on local computer to spin up minikube
+#source ./create_cluster.sh $k8s_version "virtualbox"
```
aattuluri, line 10: You can control this using a env variable like if `IS_LOCAL` == true then run this step. It would be cleaner in `Contributing.md` instructions

aattuluri (changes requested):

tests/install_istio.sh
```diff
@@ -43,10 +43,15 @@ then
...
-else
+elif [ $(ver $istio_version) -lt $(ver 1.6.0) ]
+then
     "./istio-$istio_version/bin/istioctl" manifest apply -f "istio-$istio_version/install/kubernetes/operator/examples/multicluster/values-istio-multicluster-gateways.yaml"
     #Verify that istiod is up and running
     kubectl rollout status deployment istiod -n istio-system
+else
+    "./istio-$istio_version/bin/istioctl" install -f "istio-$istio_version/manifests/examples/multicluster/values-istio-multicluster-gateways.yaml"
+    #Verify that istiod is up and running
+    kubectl rollout status deployment istiod -n istio-system
```
aattuluri, line 54: Nice 👍

aattuluri (approved):
lgtm

.circleci/config.yml
```diff
@@ -40,6 +42,52 @@ jobs:
...
+            curl -L -o /usr/local/bin/kustomize https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64
+            chmod +x /usr/local/bin/kustomize
+            kustomize version
+      - run:
+          name: Run Integration Test
+          command:  |
+            make gen-yaml
+            cd tests
+            export IS_LOCAL=false
+            ./run.sh "1.16.8" "1.5.7" "../out"
```
aattuluri, line 87: I think we should assert to make sure the tests are passing, either we can exit with status 1 in test1.sh and test2.sh or grep for PASS from the output of `./run.sh "1.16.8" "1.5.7" "../out"`. Ex: `./run.sh "1.16.8" "1.5.7" "../out" | grep "PASS" | wc -l` will return 2 since there are 2 test cases.

Also, we should separate out each combination as a separate run command. For example, we should have - run: name: `Run Integration Test for Istio 1.5.7`. That way we can see which combination failed and which passed.

aattuluri (changes requested):
Looks like the there are no assertions, please look at my comment.

.circleci/config.yml
Mengying-Li: this make sense, I'm working on it

tests/test1.sh
```diff
@@ -13,6 +13,7 @@ if [[ "$output" == *"Admiral"* ]]; then
   echo "PASS"
   exit 0
 else
-  echo "FAIL" . $output
```
aattuluri, line 16: Why is this change needed?

aattuluri (approved):
lgtm. Except for a single comment I added. Please merge it once you fix that.

tests/test1.sh
Mengying-Li: I rechecked. I'm very sure FAIL didn't get printed and I'm not very sure why... so that' why I changed it to this.

PR #146, opened 2020-10-07
Delete the SEs when deployment is deleted 

Mengying-Li:
Signed-off-by: Madeline <madeline_li@intuit.com>


admiral/pkg/clusters/serviceentry.go
```diff
@@ -554,8 +572,26 @@ func generateServiceEntry(admiralCache *AdmiralCache, meshPorts map[string]uint3
 	}
 	seEndpoint := makeRemoteEndpointForServiceEntry(endpointAddress,
 		locality, finalProtocol, port)
-	tmpSe.Endpoints = append(tmpSe.Endpoints, seEndpoint)
+
+	// if the action is deleting an endpoint from service entry, loop through the list and delete matching ones
+	if event == common.Add {
+		tmpSe.Endpoints = append(tmpSe.Endpoints, seEndpoint)
+	} else if event == common.Delete {
+		for i, exitingEndpoint := range tmpSe.Endpoints {
```
aattuluri, line 580: There is typo, can you refactor to `existingEndpoint`

admiral/pkg/clusters/serviceentry.go
```diff
@@ -554,8 +572,26 @@ func generateServiceEntry(admiralCache *AdmiralCache, meshPorts map[string]uint3
...
+			if reflect.DeepEqual(exitingEndpoint, seEndpoint) {
+				tmpSe.Endpoints = RemoveIndex(tmpSe.Endpoints, i)
```
aattuluri, line 582: Is this not causing index problems as you iterate and remove elements from the same list? why not creating another with filtered endpoints and swap it at the end. Definitely add a test case for this.

admiral/pkg/clusters/types.go
```diff
@@ -391,14 +396,36 @@ func (pc *DeploymentHandler) Added(obj *k8sAppsV1.Deployment) {
...
+		log.Infof(LogFormat, "Event", "deployment", obj.Name, "", "Skipped as '"+common.GetWorkloadIdentifier()+" was not found', namespace="+obj.Namespace)
+		return
+	}
+
+	var matchedGTPs []v1.GlobalTrafficPolicy
+	for _, remoteCluster := range pc.RemoteRegistry.remoteControllers {
+		matchedGTPs = append(matchedGTPs, remoteCluster.GlobalTraffic.GetGTPByLabel(obj.Labels[common.GetGlobalTrafficDeploymentLabel()], obj.Namespace)...)
+	}
+
+	gtp := common.MatchGTPsToDeployment(matchedGTPs, obj)
```
aattuluri, line 417: Refactor this code like we talked about.

aattuluri (changes requested):

admiral/pkg/clusters/serviceentry.go
Mengying-Li: yep, I made a mistake here I think it might caused the problem you said

aattuluri (approved):

Issue #76, opened 2020-02-29
[BUG] Infer ServiceEntry ports from destination service


josephpeacock:
The serviceEntry should add ports based on those present in the service it will be pointing to. Currently it is hardcoded to default http ports (and named http).

Thanks to Martin Baillie for the report!

> for those at Intuit more familiar with the code, I was thinking these lines should not be hardcoded and should instead be inferred from the respective Service  ports and port names?
https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/clusters/serviceentry.go#L84-L85
since my Service ports are named grpc I am being bitten by: https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry-Endpoint

aattuluri:
Fixed with https://github.com/istio-ecosystem/admiral/pull/148

PR #148, opened 2020-10-29
Infer protocol from k8s service name

aattuluri:
Context:
https://github.com/istio-ecosystem/admiral/issues/76

This adds support for creating admiral endpoints with grpc as the protocol.

`Note:` Currently only one of http or grpc protocols is supported for a given service because there is only one port for cross cluster communication which is 15443. This support will be expanded upon auto protocol detection across clusters on the same port.


josephpeacock (changes requested):
A few minor comments in-line. The big one is I see a few `if len(meshPorts) > 1 {` blocks, but it's not everywhere that the meshPorts is used. Let's either a) make sure we're checking that error case everywhere or b) check it in the GetMeshPorts method

More broadly, is doing nothing the right approach for a situation in which a service has multiple ports? I don't feel confident that I know the right answer here, but failing silently (even with logs) feels like an invitation for lots of future debugging time in response to the question "Why aren't I seeing a service entry for my new deployment?" Feels like something we should at least make an intentional decision on.

josephpeacock:
@aattuluri One more thought - I see you added examples to the sample yamls, do you want to also add the accompanying instructions here https://github.com/istio-ecosystem/admiral/blob/master/docs/Examples.md?

aattuluri:
> A few minor comments in-line. The big one is I see a few `if len(meshPorts) > 1 {` blocks, but it's not everywhere that the meshPorts is used. Let's either a) make sure we're checking that error case everywhere or b) check it in the GetMeshPorts method
> 
> More broadly, is doing nothing the right approach for a situation in which a service has multiple ports? I don't feel confident that I know the right answer here, but failing silently (even with logs) feels like an invitation for lots of future debugging time in response to the question "Why aren't I seeing a service entry for my new deployment?" Feels like something we should at least make an intentional decision on.

Valid point, this made me thinking into keeping the behavior backward compatible and also consistent. This approach will pick the first port/protocol when more than one is present and logs a warning if that multi port support isn't there.

josephpeacock:
I think that makes a lot of sense.

josephpeacock (approved):
lgtm

josephpeacock:
You'll have to go fix the sign-off. But once that's done and the build passes, I'd say it looks good!

PR #153, opened 2020-11-13
Use admiral.io/env as the primary way to group deployments by env

aattuluri:


admiral/pkg/controller/common/common.go
```diff
@@ -88,15 +88,19 @@ func GetCname(deployment *k8sAppsV1.Deployment, identifier string, nameSuffix st
...
+	}
+	if len(environment) == 0 {
+		environment = deployment.Spec.Template.Labels[Env]
 	}
 	if len(environment) == 0 {
 		splitNamespace := strings.Split(deployment.Namespace, Dash)
 		if len(splitNamespace) > 1 {
 			environment = splitNamespace[len(splitNamespace)-1]
 		}
+		log.Infof("Using deprecated approach to deduce env from namespace for deployment, name=%v in namespace=%v", deployment.Name, deployment.Namespace)
```
josephpeacock, line 103: Should this be a warning?

admiral/cmd/admiral/cmd/root.go
```diff
@@ -93,6 +93,8 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The parameter will be used to decide whether to update workload sidecar resource or not. By default these updates will be disabled.")
 	rootCmd.PersistentFlags().StringVar(&params.WorkloadSidecarName, "workload_sidecar_name", "default",
 		"Name of the sidecar resource in the workload namespace. By default sidecar resource will be named as \"default\".")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.EnvKey, "env_key", "admiral.io/env",
```
josephpeacock, line 96: Should we be more specific on behavior here? The behavior seems to be looking at (in order) annotation with envKey, label with envKey, label with env, pulling from ns. Saying that directly will prevent someone from needing to go through the code to learn it.

admiral/pkg/controller/common/types.go
```diff
@@ -64,6 +64,7 @@ type LabelSet struct {
 	AdmiralIgnoreLabel                  string
 	WorkloadIdentityKey                 string //Should always be used for both label and annotation (using label as the primary, and falling back to annotation if the label is not found)
 	GlobalTrafficDeploymentLabel        string //label used to tie together deployments and globaltrafficpolicy objects. Configured separately from the identity key because this one _must_ be a label
+	EnvKey								string //should be used for annotation, annotation is primary and label (env) as a fallback
```
josephpeacock, line 67: Please update this comment to also specify that we check annotation, then label, then env label

josephpeacock (changes requested):
Functionality looks good. Mostly comments around documentation/logging.

Two other notes:
- Should we add this annotation to the sample?
- Please update the docs to reflect the change (I know this file explicitly references the env label, not sure if others https://github.com/istio-ecosystem/admiral/blob/master/docs/Architecture.md)

josephpeacock (approved):
lgtm

PR #162, opened 2021-01-04
Add support for gtp with multiple traffic policies

aattuluri:


admiral/pkg/clusters/handler.go
```diff
@@ -176,16 +176,25 @@ func makeVirtualService(host string, gateways []string, destination string, port
...
 	var dr = &v1alpha32.DestinationRule{}
 	dr.Host = host
 	dr.TrafficPolicy = &v1alpha32.TrafficPolicy{Tls: &v1alpha32.TLSSettings{Mode: v1alpha32.TLSSettings_ISTIO_MUTUAL}}
-	if gtpWrapper != nil {
+	processGtp := true
+	if len(locality) == 0 {
+		log.Errorf(LogErrFormat, "Process", "GlobalTrafficPolicy", host, "", "Skipping gtp processing, locality of the cluster nodes cannot be determined. Is this minikube?")
+		processGtp = false
+	}
+	outlierDetection := &v1alpha32.OutlierDetection{
```
josephpeacock, line 188: This may be out of scope for this PR, but is there any plans to make these values configurable in any way? If not, are we sure these are the defaults we want? (100 consecutive errors seems high to me)

admiral/pkg/clusters/serviceentry.go
```diff
@@ -24,6 +25,13 @@ import (
 	"github.com/istio-ecosystem/admiral/admiral/pkg/controller/util"
 )
 
+type SeDrTuple struct {
+	SeName string
```
josephpeacock, line 29: do the SeName and DrName need to be separate fields in the struct? Since they will also be accessible via ServiceEntry.Name, etc.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -288,14 +296,21 @@ func createSeWithDrLabels(remoteController *RemoteController, localCluster bool,
...
 }
 
+//This will create the default service entries and also additional ones specified in GTP
 func AddServiceEntriesWithDr(cache *AdmiralCache, sourceClusters map[string]string, rcs map[string]*RemoteController, serviceEntries map[string]*networking.ServiceEntry) {
 	syncNamespace := common.GetSyncNamespace()
 	for _, se := range serviceEntries {
 
-		//add service entry
-		serviceEntryName := getIstioResourceName(se.Hosts[0], "-se")
+		//Add a label
```
josephpeacock, line 304: Is this comment accurate?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -306,46 +321,85 @@ func AddServiceEntriesWithDr(cache *AdmiralCache, sourceClusters map[string]stri
...
+				oldServiceEntry, err := rc.ServiceEntryController.IstioClient.NetworkingV1alpha3().ServiceEntries(syncNamespace).Get(seDr.SeName, v12.GetOptions{})
+				// if old service entry not find, just create a new service entry instead
+				if err != nil {
+					log.Infof(LogFormat, "Get (error)", "old ServiceEntry", seDr.SeName, sourceCluster, err)
+					oldServiceEntry = nil
+				}
+				oldDestinationRule, err := rc.DestinationRuleController.IstioClient.NetworkingV1alpha3().DestinationRules(syncNamespace).Get(seDr.DrName, v12.GetOptions{})
 
-			oldDestinationRule, err := rc.DestinationRuleController.IstioClient.NetworkingV1alpha3().DestinationRules(syncNamespace).Get(destinationRuleName, v12.GetOptions{})
+				if err != nil {
```
josephpeacock, line 335: Should we be logging this error?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -306,46 +321,85 @@ func AddServiceEntriesWithDr(cache *AdmiralCache, sourceClusters map[string]stri
...
+	var defaultDrName = se.Hosts[0] + "-default-dr"
+	var defaultSeName = se.Hosts[0] + "-se"
+	var seDrSet = make(map[string]*SeDrTuple)
+	if globalTrafficPolicy != nil {
+		gtp := globalTrafficPolicy.Spec
+		for _, gtpTrafficPolicy := range gtp.Policy {
+			var modifiedSe *networking.ServiceEntry
+			var host = se.Hosts[0]
+			var drName, seName = defaultDrName, defaultSeName
+			if gtpTrafficPolicy.Dns != env && gtpTrafficPolicy.Dns != common.Default {
```
josephpeacock, line 372: Do you possible mean some other comparison here? Is it possible that the DNS = env?

install/sample/gtp.yaml
```diff
@@ -3,11 +3,13 @@ kind: GlobalTrafficPolicy
...
   name: gtp-service1
   namespace: sample
+  annotations:
+    admiral.io/env: stage
   labels:
     identity: greeting
 spec:
   policy:
-    - dns: stage.greeting.global
+    - dns: default # default is a keyword, alternatively you can use `env` (ex: stage)
```
josephpeacock, line 12: Connecting to my previous comment ```Do you possible mean some other comparison here? Is it possible that the DNS = env?``` - could we maybe document this behavior a little bit more clearly? I find it a little counterintuitive that the DNS field should be the relevant environment (and it's distinctly different from how the docs say things work: https://github.com/istio-ecosystem/admiral/blob/master/docs/Architecture.md)

josephpeacock (changes requested):

josephpeacock:
The change from the DNS field in the GTP to be the environment seems to be backwards incompatible to me. Are we concerned that may break any existing users?

.circleci/config.yml
```diff
@@ -90,18 +90,18 @@ jobs:
...
             ./run.sh "1.16.8" "1.5.7" "../out"
+#      - run:
+#          name: Run Integration Test for Istio 1.6.13
+#          command:  |
+#            cd tests
+#            export IS_LOCAL=false
+#            ./run.sh "1.16.8" "1.6.13" "../out"
       - run:
-          name: Run Integration Test for Istio 1.6.13
+          name: Run Integration Test for Istio 1.7.6
```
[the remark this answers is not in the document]
aattuluri: Good point, The script `https://istio.io/downloadIstioctl` doesn't have this capability and needs to be tweaked, will file another issue to get this in.

aattuluri:
> The change from the DNS field in the GTP to be the environment seems to be backwards incompatible to me. Are we concerned that may break any existing users?

I changed the implementation to treat dns as a dns prefix rather than FQQN, the rationale is as follows:  
Currently, we generate default service entries based on the env label and they look something like <env>.<identity>.global. Now when a gtp is defined, repeating the fully qualified domain name for every dns is error prone and cannot guarantee uniqueness across envs and namespaces. The dns field will act as a dnsPrefix except for scenarios where dns = 'default' (a keyword) or dns = `env` in which case its assumed that someone is trying to define override for locality load balancing via gtp for the default name. This will actually preserve backward compatibility and make sure the generated entries are unique. 

That being said, I can add an additional check to see if the dns in gtp matches the full default host for backward compatibility but that will retain hybrid behavior (dns prefix vs full dns), so I think it makes sense to introduce this backward incompatible change and then call it out in the release notes.

admiral/pkg/clusters/serviceentry.go
aattuluri: ServiceEntry and DestinationRule in this struct are istio model objects and not k8s objects, so the `Name` field isn't available.

admiral/pkg/clusters/handler.go
aattuluri: Good point, the increase to 100 was more to fix the flickery behavior upon 503s but this is backward incompatible behavior change. I will switch it back to 10 until we make this configurable (a separate issue exists)

josephpeacock:
I agree 100% that this is a good direction in which to go. I just want to make sure we're not breaking anyone else. Not sure how careeful we want to be, but the most foolproof option would probably be to introduce a new field in the GTP spec (something like `DNSPrefix`) and mark `DNS` as deprecated.

josephpeacock (approved):
Approving as all of your updates look good, but please update the GTP section here: https://github.com/istio-ecosystem/admiral/blob/master/docs/Architecture.md to use the DNS prefix and show how it's used.

Issue #164, opened 2021-01-21
[BUG] k8s service's Target port and Port being different causes wrong port to be picked up


aattuluri:
**Describe the bug**
If a k8s service's target port is different from the port field then the right port is not picked up

**Steps To Reproduce**
Having the following objects will help reproduce the error:

k8s service:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app: greeting
    identity: greeting
  name: greeting
spec:
  ports:
    - port: 80
      targetPort: 8080
      name: http
  selector:
    app: greeting
```

k8s Deployment:
```
...
  spec:
      template:
         annotations:
              traffic.sidecar.istio.io/includeInboundPorts: "8080"
...
```

The above configuration doesn't result in a port match causing invalid service entry to be generated.

**Expected behavior**

aattuluri:
Fixed with https://github.com/istio-ecosystem/admiral/pull/165

PR #165, opened 2021-01-21
Fix mesh port match against to look at k8s svc target port

aattuluri:
Context: https://github.com/istio-ecosystem/admiral/issues/164


josephpeacock (approved):
lgtm

PR #175, opened 2021-04-30
Admiral apis

vrushalijoshi:
Admiral APIs - 
1. Get all clusters admiral is watching
2. Get all se by cluster name
3. Get all se by service identity

### DIFF
--- a/admiral/pkg/apis/admiral/routes/handlers.go
+++ b/admiral/pkg/apis/admiral/routes/handlers.go
@@ -24,5 +24,18 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request) 
 }
 
 func (opts *RouteOpts) GetClusters(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(200)
+	clusterList := ""
+	// loop through secret controller's c.cs.remoteClusters to access all clusters admiral is watching
+	for clusterID := range opts.RemoteRegistry.SecretController.Cs.RemoteClusters {
+		fmt.Print(clusterID)
+		clusterList += " " + clusterID
+	}
+	response := fmt.Sprintf(clusterList)
 
+	_, writeErr := w.Write([]byte(response))
+	if writeErr != nil {
+		log.Printf("Error writing body: %v", writeErr)
+		http.Error(w, "can't write body", http.StatusInternalServerError)
+	}
 }
--- a/admiral/pkg/apis/admiral/routes/handlers.go
+++ b/admiral/pkg/apis/admiral/routes/handlers.go
@@ -21,8 +21,8 @@ type RouteOpts struct {
 //}
 
 type IdentityServiceEntry struct {
-	IdentityCname string   `json:"IdentityCname,omitempty"`
-	ClusterNames  []string `json:"Clusters,omitempty"`
+	Cname        string   `json:"Cname,omitempty"`
+	ClusterNames []string `json:"Clusters,omitempty"`
 }
 
 func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request) {
@@ -38,7 +38,7 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request) 
 
 func (opts *RouteOpts) GetClusters(w http.ResponseWriter, r *http.Request) {
 
-	var clusterList []string
+	clusterList := []string{}
 
 	// loop through secret controller's c.cs.remoteClusters to access all clusters admiral is watching
 	for clusterID := range opts.RemoteRegistry.SecretController.Cs.RemoteClusters {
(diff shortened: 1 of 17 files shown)


admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -25,4 +38,94 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request)
...
+
+		serviceEntriesByCluster, err := clusters.GetServiceEntriesByCluster(clusterName, opts.RemoteRegistry)
+
+		if err != nil {
+			log.Printf(err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		} else {
+			if len(serviceEntriesByCluster) == 0 {
+				log.Printf(fmt.Sprintf("No service entries configured for cluster - %s", clusterName))
+				http.Error(w, fmt.Sprintf("No service entries configured for cluster - %s", clusterName), http.StatusNotFound)
```
aattuluri, line 77: I think the behavior should be:
The cluster doesn't exist (is not being monitored by admiral) -> 404
The cluster has 0 service entries -> empty response object with 200

admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -25,4 +38,94 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request)
...
+			} else {
+				response = serviceEntriesByCluster
+
+				out, err := json.Marshal(response)
+				if err != nil {
+					log.Printf("Failed to marshall response")
```
aattuluri, line 83: Same as above, provide request context.

aattuluri (changes requested):
Good work!
Please see comments and add unit tests please.

admiral/pkg/apis/admiral/routes/handlers.go
vrushalijoshi: Adding fix for this

admiral/pkg/apis/admiral/routes/handlers.go
vrushalijoshi: Makes sense, updated the response handling

admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -25,4 +38,94 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request)
...
+	var clusterList []string
+
+	// loop through secret controller's c.cs.remoteClusters to access all clusters admiral is watching
+	for clusterID := range opts.RemoteRegistry.SecretController.Cs.RemoteClusters {
+		clusterList = append(clusterList, clusterID)
+	}
+
+	out, err := json.Marshal(clusterList)
+	if err != nil {
+		log.Printf("Failed to marshall response")
```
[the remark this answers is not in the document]
vrushalijoshi: done

admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -1,17 +1,30 @@
...
 	KubeconfigPath string
 	RemoteRegistry *clusters.RemoteRegistry
 }
 
+//type ClusterServiceEntries struct {
+//	ServiceEntries []v1alpha3.ServiceEntry `json:"ServiceEntries,omitempty"`
+//}
+
+type IdentityServiceEntry struct {
+	IdentityCname string   `json:"IdentityCname,omitempty"`
```
aattuluri, line 24: Can this be changed to just Cname?, IdentityCname seems a bit misleading

aattuluri (approved):

admiral/pkg/apis/admiral/routes/handlers.go
vrushalijoshi: Sure, added fix for this

Mengying-Li:
@vrushalijoshi  just a quick question. I'm trying to understand why the SeClusterCache needs to be a map of map. I notice this pattern many times in other caches as well. But why do we need to put host, clusterID, clusterID this format? Why twice on the cluster ID? This part is just a question for my own understanding.

aattuluri:
> @vrushalijoshi just a quick question. I'm trying to understand why the SeClusterCache needs to be a map of map. I notice this pattern many times in other caches as well. But why do we need to put host, clusterID, clusterID this format? Why twice on the cluster ID? This part is just a question for my own understanding.

Because go doesn't support hash set, we could have simply done host, clusterID, true kind of format as well

Mengying-Li:
> Because go doesn't support hash set, we could have simply done host, clusterID, true kind of format as well

why do we need true here? Doesn't host, clusterId and delete when not used is good enough? @aattuluri

admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -25,4 +38,106 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request)
...
+	clusterName := strings.Trim(params["clustername"], " ")
+
+	var response []v1alpha3.ServiceEntry
+
+	if clusterName != "" {
+
+		serviceEntriesByCluster, err := clusters.GetServiceEntriesByCluster(clusterName, opts.RemoteRegistry)
+
+		if err != nil {
+			log.Printf(err.Error())
```
aattuluri, line 80: Print some request context, would be hard to understand what this error is for

admiral/pkg/apis/admiral/routes/handlers.go
```diff
@@ -25,4 +38,106 @@ func (opts *RouteOpts) ReturnSuccessGET(w http.ResponseWriter, r *http.Request)
...
+			if strings.Contains(err.Error(), "Admiral is not monitoring cluster") {
+				http.Error(w, err.Error(), http.StatusNotFound)
+			} else {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+		} else {
+			if len(serviceEntriesByCluster) == 0 {
+				log.Printf(fmt.Sprintf("No service entries configured for cluster - %s", clusterName))
```
aattuluri, line 88: Print some request context

aattuluri (changes requested):

aattuluri (approved):

admiral/pkg/apis/admiral/routes/handlers.go
Mengying-Li: Sorry I just saw the comment. I will address that in the other branch @aattuluri

PR #176, opened 2021-05-08
added the cluster id inside deployment controller

Mengying-Li:
Signed-off-by: Mengying <mengyinglimandy@gmail.com>

### DIFF
--- a/admiral/pkg/clusters/registry.go
+++ b/admiral/pkg/clusters/registry.go
@@ -129,7 +129,7 @@ func (r *RemoteRegistry) createCacheController(clientConfig *rest.Config, cluste
 	}
 
 	log.Infof("starting deployment controller clusterID: %v", clusterID)
-	rc.DeploymentController, err = admiral.NewDeploymentController(stop, &DeploymentHandler{RemoteRegistry: r}, clientConfig, resyncPeriod)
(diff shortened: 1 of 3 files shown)


admiral/pkg/clusters/registry.go
```diff
@@ -129,7 +129,7 @@ func (r *RemoteRegistry) createCacheController(clientConfig *rest.Config, cluste
 	}
 
 	log.Infof("starting deployment controller clusterID: %v", clusterID)
-	rc.DeploymentController, err = admiral.NewDeploymentController(stop, &DeploymentHandler{RemoteRegistry: r}, clientConfig, resyncPeriod)
+	rc.DeploymentController, err = admiral.NewDeploymentController(stop, &DeploymentHandler{RemoteRegistry: r, ClusterID: clusterID}, clientConfig, resyncPeriod)
```
aattuluri, line 132: Are there other controllers missing ClusterId?

aattuluri, line 132: Are there other controllers missing ClusterId?

aattuluri (changes requested):

admiral/pkg/clusters/registry.go
Mengying-Li: Would you like me to update all of them ? @aattuluri

aattuluri: Yep, that one is fine.

PR #180, opened 2021-06-17
Add latest istio release support

aattuluri:


asushanthk (approved):
Looks good.

admiral/pkg/clusters/handler.go
```diff
@@ -577,16 +476,6 @@ func addUpdateVirtualService(obj *v1alpha3.VirtualService, exist *v1alpha3.Virtu
 	}
 }
 
-func deleteVirtualService(exist *v1alpha3.VirtualService, namespace string, rc *RemoteController) {
```
vrushalijoshi, line 580: Will cleanup of virtual service handled in a different way or we need to have this to take care of it

vrushalijoshi (approved):
Looks good, added a comment for clarification

admiral/pkg/clusters/handler.go
aattuluri: We no longer create virtual services (as they were redundant since Istio 1.6 as I recall for cross cluster traffic), so delete isn't required either.

vrushalijoshi: ok got it. Makes sense

PR #185, opened 2021-11-18
Remove Pod Controller

frankmariette:
The Pod controller uses a substantial amount of memory in large clusters though isn't being utilized in Admiral. Removing this drops memory usage requirements drastically.


.gitignore
```diff
@@ -2,5 +2,6 @@
 .idea/vcs.xml
 
 out*
-
-c.out
\ No newline at end of file
+*.out
+istio-*/
```
aattuluri, line 6: Remove these?

aattuluri (changes requested):

.gitignore
frankmariette: Sure. Had added these here as they are artifacts from the setup scripts though can remove them as they aren't needed for this PR.

aattuluri (approved):
lgtm

PR #186, opened 2021-11-18
Fix CircleCI build and Golang linting

frankmariette:
CI builds are failing due to the certificate change shown [here](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/). Updating to the latest image from Circle CI fixes this problem and allows Admiral to build again as expected. 

Fixed linter to run though  did not make changes to any current files to limit this to just fixing CI and not lint issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,6 +54,7 @@ jobs:
       MINIKUBE_WANTREPORTERRORPROMPT: false
       MINIKUBE_HOME: /home/circleci
       CHANGE_MINIKUBE_NONE_USER: true
+    resource_class: large
     steps:
       - attach_workspace:
           at: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     working_directory: /go/pkg/mod/github.com/admiral
     docker:
-    - image: circleci/golang:1.11
+    - image: circleci/golang:1.16
     steps:
     - checkout
     - run:
@@ -104,7 +104,7 @@ jobs:
             ./run.sh "1.16.8" "1.7.6" "../out"
   publish-github-release:
     docker:
-    - image: circleci/golang:1.11
+    - image: circleci/golang:1.16
     working_directory: /go/pkg/mod/github.com/admiral
     steps:
     - attach_workspace:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   build:
     working_directory: /go/pkg/mod/github.com/admiral
     docker:
-    - image: circleci/golang:1.11
+    - image: circleci/golang:1.16
```
nirvanagit, line 10: Should we use 1.17 ?

.circleci/config.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       MINIKUBE_WANTREPORTERRORPROMPT: false
       MINIKUBE_HOME: /home/circleci
       CHANGE_MINIKUBE_NONE_USER: true
+    resource_class: large
```
nirvanagit, line 57: Curious to know why we need a large build container (assuming it is a build container)

tests/create_cluster.sh
```diff
@@ -5,7 +5,7 @@
 k8s_version=$1
 
 if [[ $IS_LOCAL == "false" ]]; then
-  sudo -E minikube start --vm-driver=none --cpus 4 --memory 4096 --kubernetes-version=$k8s_version &> $HOME/minikube.log 2>&1 < /dev/null
+  sudo -E minikube start --vm-driver=none --cpus 6 --memory 6144 --kubernetes-version=$k8s_version &> $HOME/minikube.log 2>&1 < /dev/null
```
nirvanagit, line 8: Is the increment in resources due to some new dependency?

frankmariette: Increased demands for the tests with added GRPC clients. I noticed the tests were flaky and seemed to be failing to timeout. Experience with running minikube has shown that the lack of responsiveness occurs when it becomes resource strapped so I increased it to account for this. I also had to increase the integration test vm resources to account for this change as well.

.circleci/config.yml
frankmariette: This is for the integration tests. The container was increased to match the size that CircleCI has with the limits I increased minikube to. https://circleci.com/docs/2.0/configuration-reference/#resourceclass

aattuluri (approved):

PR #191, opened 2022-02-02
Sync VirtualServices and DestinationRules `as is` to other clusters if not generated by Admiral

aattuluri:


vrushalijoshi (approved):
Looks good

PR #197, opened 2022-03-10
fix for concurrent map iteration and map write

### LINKED ISSUE #196: [BUG] concurrent map iteration and map write
# Current Issue:
There are instances in the code where we iterating over a map while another go routine is simultaneously trying to write to it. 

https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/apis/admiral/routes/handlers.go#L129-L133
https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/clusters/handler.go#L721-L724

# Proposed Fix:
The fix would be to introduce a `.Range()` func, similar to `sync.Map`, which can enable a thread-safe way of iterating through the map.

shriramsharma:
fixes #196

Added `.Range()` and `.Copy()` func to avoid concurrent map read and
write.


aattuluri (approved):
lgtm, please fix the lint errors.

PR #199, opened 2022-03-18
Support for BlueGreen rollout preview endpoint

itsLucario:
Currently when Admiral generates endpoint for Rollouts with bluegreen it always uses active/stable service as destination. This PR will adds a feature to generate an additional mesh endpoint that would point to the preview service of the blugreen stack.


admiral/pkg/clusters/handler.go
```diff
@@ -844,7 +847,7 @@ func getServiceForRollout(rc *RemoteController, rollout *argo.Rollout) map[strin
 		var service = servicesInNamespace[s]
 		var match = true
 		//skip services that are not referenced in the rollout
-		if len(blueGreenActiveService) > 0 && service.ObjectMeta.Name != blueGreenActiveService {
+		if len(blueGreenActiveService) > 0 && len(blueGreenPreviewService) > 0 && service.ObjectMeta.Name != blueGreenActiveService && service.ObjectMeta.Name != blueGreenPreviewService {
```
asushanthk, line 850: blueGreenPreviewService can be null

admiral/cmd/admiral/cmd/root.go
```diff
@@ -103,6 +104,8 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The label value, on a namespace, which tells Istio to perform sidecar injection")
 	rootCmd.PersistentFlags().StringVar(&params.HostnameSuffix, "hostname_suffix", "global",
 		"The hostname suffix to customize the cname generated by admiral. Default suffix value will be \"global\"")
+	rootCmd.PersistentFlags().StringVar(&params.PreviewHostnamePrefix, "preview_hostname_prefix", "preview",
```
aattuluri, line 107: I dont think we need to make it configurable, lets just use `preview` the standard name for bluegreen strategy.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -132,12 +134,17 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
...
 		var meshPorts map[string]uint32
+		isBlueGreenStrategy := false
+
+		if len(sourceRollouts) > 0 {
+			isBlueGreenStrategy = sourceRollouts[sourceCluster].Spec.Strategy.BlueGreen != nil
+		}
+
 		if len(sourceDeployments) > 0 {
 			meshPorts = GetMeshPorts(sourceCluster, serviceInstance, sourceDeployments[sourceCluster])
-		} else {
```
aattuluri, line 137: why not return active and preview services both as a response, that way we dont have to make this custom. You can then handle that with a method similar to `updateEndpointsForWeightedServices`?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -147,8 +154,31 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
 			for _, ep := range serviceEntry.Endpoints {
 				//replace istio ingress-gateway address with local fqdn, note that ingress-gateway can be empty (not provisoned, or is not up)
 				if ep.Address == clusterIngress || ep.Address == "" {
-					// see if we have weighted services (rollouts with canary strategy)
-					if len(weightedServices) > 1 {
+					// Update endpoints with locafqdn for active and preview se of bluegreen rollout
+					if isBlueGreenStrategy {
+						activeServiceName := sourceRollouts[sourceCluster].Spec.Strategy.BlueGreen.ActiveService
```
aattuluri, line 159: Abstract all this logic into a method similar to `updateEndpointsForWeightedServices`

aattuluri (changes requested):

admiral/pkg/clusters/serviceentry.go
```diff
@@ -179,6 +196,26 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
...
+	activeServiceName := rollout.Spec.Strategy.BlueGreen.ActiveService
+	previewServiceName := rollout.Spec.Strategy.BlueGreen.PreviewService
+
+	if previewService, ok := weightedServices[previewServiceName]; strings.HasPrefix(meshHost, common.BlueGreenRolloutPreviewPrefix+common.Sep) && ok {
+		previewServiceInstance := previewService.Service
+		localFqdn := previewServiceInstance.Name + common.Sep + previewServiceInstance.Namespace + common.DotLocalDomainSuffix
+		cnames[localFqdn] = "1"
+		ep.Address = localFqdn
+		ep.Ports = GetMeshPortsForRollout(sourceCluster, previewServiceInstance, rollout)
+	} else if activeService, ok := weightedServices[activeServiceName]; ok {
```
aattuluri, line 210: Shouldn't this be `if` instead of `else if`, will this method (updateEndpointsForBlueGreen) be called twice once for active and preview endpoints?

aattuluri (changes requested):
lgtm, added a question

admiral/pkg/clusters/serviceentry.go
itsLucario: Yes, this will be called twice, once for active and preview endpoints.

itsLucario: If `meshHost` is a preview host then `else if` should not execute. Making it as `if` will overwrite with active service address to preview endpoint.

aattuluri (approved):
lgtm

aattuluri:
@itsLucario There are lint errors, fix them before we can merge.

itsLucario:
> @itsLucario There are lint errors, fix them before we can merge.

golangci-lint was using go1.18 that caused the panic. I added the fix to the workflow to use go1.17.7.

PR #208, opened 2022-04-26
Prometheus Integration

adilfulara:
**Objective**

- Prometheus integration
- - exposed on port 6900
- Gauge for clusters monitored by Admiral
- - metric name is: clusters_monitored


admiral/cmd/admiral/cmd/root.go
```diff
@@ -116,6 +122,7 @@ func GetRootCmd(args []string) *cobra.Command {
 			"The order would be to use annotation specified as `env_key`, followed by label specified as `env_key` and then fallback to the label `env`")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GatewayApp, "gateway_app", "istio-ingressgateway",
 		"The the value of the `app` label to use to match and find the service that represents the ingress for cross cluster traffic (AUTO_PASSTHROUGH mode)")
+	rootCmd.PersistentFlags().BoolVar(&params.PrometheusEnabled, "prometheus", true, "Enable prometheus metrics collections")
```
aattuluri, line 125: Rename `MetricsEnabled` and `metrics`. You can leave the comment as is.

admiral/cmd/admiral/cmd/root.go
```diff
@@ -50,14 +50,20 @@ func GetRootCmd(args []string) *cobra.Command {
...
 
 			if err != nil {
 				log.Error("Error setting up server:", err.Error())
 			}
 
-			service.Start(ctx, 8080, ret_routes, routes.Filter, remoteRegistry)
+			go func() {
+				metricsService.Start(ctx, 6900, metricRoutes, routes.Filter, remoteRegistry)
+			}()
+			service.Start(ctx, 8080, mainRoutes, routes.Filter, remoteRegistry)
```
aattuluri, line 66: This server should be started only if metrics are enabled?

admiral/pkg/controller/secret/secretcontroller.go
```diff
@@ -45,6 +45,8 @@ const (
 // DO NOT USE - TEST ONLY.
 var LoadKubeConfig = clientcmd.Load
 
+var remoteClustersMetric common.Gauge
```
aattuluri, line 48: You don't need an instance, you simply use a method like `common.SetGuage(...)` and in SetGuage you can initialize if one doesn't exist and update the guage?

aattuluri (changes requested):

admiral/cmd/admiral/cmd/root.go
adilfulara: currently implementation has 2 options. prometheus and noop. that way a 200 is always returned but the noop returns no data.

admiral/pkg/controller/secret/secretcontroller.go
adilfulara: common.Guage is a thin wrapper over a implementation specific gauge ( prom / noop ) which is why i wrote it as an instance. in addition, i added interface so behavior of gauge is modeled.

admiral/cmd/admiral/cmd/root.go
adilfulara: done.

aattuluri (approved):
lgtm

PR #210, opened 2022-05-02
[bug] prevent panic when finding out rollout strategy

nirvanagit:
## Description
Argo rollouts can be defined without any rollout strategy. When this happens admiral panics because the code was fetch the Strategy when Strategy was empty.


aattuluri (approved):
lgtm

PR #220, opened 2022-05-20
Added mutex to sync access to the SidecarEgressMap

### LINKED ISSUE #219: [BUG]: SidecarEgressMap.Get() throws concurrent map access error intermittently
**Describe the bug**
SidecarEgressMap.Get() throws concurrent map access error intermittently.
The map access inside the func is not thread-safe. 
https://github.com/istio-ecosystem/admiral/blob/72d5ec46a610b4c94ca78d5148507cb1387ceb4a/admiral/pkg/controller/common/types.go#L179

**Steps To Reproduce**
This error is pretty intermittent. I wrote a test to reproduce this error.
```
func TestSidecarEgressGet(t *testing.T) {

	egressMap := NewSidecarEgressMap()
	egressMap.Put("pkey1", "pkey2", "fqdn", map[string]string{"pkey2": "pkey2"})

	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(5*time.Second))
	defer cancel()

	go func(ctx context.Context) {
		for {
			select {
			case <-ctx.Done():
				return
			default:
				egressMap.Put("pkey1", string(uuid.NewUUID()), "fqdn", map[string]string{"pkey2": "pkey2"})
			}
		}
	}(ctx)

	for {
		select {
		case <-ctx.Done():
			return
		default:
			assert.NotNil(t, egressMap.Get("pkey1"))
		}
	}

}
```
Error:
```
fatal error: concurrent map read and map write

goroutine 10 [running]:
runtime.throw({0x181ef9c, 0xc000495980})
	/Users/ssharma34/shriram/sandbox/goroots/go1.17/src/runtime/panic.go:1198 +0x71 fp=0xc000083e70 sp=0xc000083e40 pc=0x1036df1
runtime.mapaccess1_faststr(0xc0001127e0, 0xc000203a00, {0x180b471, 0x5})
	/Users/ssharma34/shriram/sandbox/goroots/go1.17/src/runtime/map_faststr.go:21 +0x3a5 fp=0xc000083ed8 sp=0xc000083e70 pc=0x1014045
github.com/istio-ecosystem/admiral/admiral/pkg/controller/common.(*SidecarEgressMap).Get(...)
	/Users/ssharma34/shriram/personal/github/admiral/admiral/pkg/controller/common/types.go:180
```

**Expected behavior**
Have a mutex around the map for synchronized access.

shriramsharma:
Fixes #219
Added mutex to sync access to the SidecarEgressMap

### DIFF
--- a/admiral/pkg/controller/common/types.go
+++ b/admiral/pkg/controller/common/types.go
@@ -177,10 +177,9 @@ func (s *SidecarEgressMap) Put(identity string, namespace string, fqdn string, c
 }
 
 func (s *SidecarEgressMap) Get(key string) map[string]SidecarEgress {
+	defer s.mutex.Unlock()
 	s.mutex.Lock()
-	val := s.cache[key]
-	s.mutex.Unlock()
-	return val
+	return s.cache[key]
 }
 
 func (s *SidecarEgressMap) Delete(key string) {
@@ -198,9 +197,9 @@ func (s *SidecarEgressMap) Map() map[string]map[string]SidecarEgress {
 
 // Range is a thread safe iterator to iterate through the SidecarEgress map
 func (s *SidecarEgressMap) Range(fn func(k string, v map[string]SidecarEgress)) {
+	defer s.mutex.Unlock()
 	s.mutex.Lock()
 	for k, v := range s.cache {
 		fn(k, v)
 	}
-	s.mutex.Unlock()
 }
--- a/admiral/pkg/controller/common/types.go
+++ b/admiral/pkg/controller/common/types.go
@@ -177,7 +177,10 @@ func (s *SidecarEgressMap) Put(identity string, namespace string, fqdn string, c
 }
 
 func (s *SidecarEgressMap) Get(key string) map[string]SidecarEgress {
-	return s.cache[key]
+	s.mutex.Lock()
+	val := s.cache[key]
+	s.mutex.Unlock()
+	return val
 }
 
 func (s *SidecarEgressMap) Delete(key string) {
@@ -186,6 +189,18 @@ func (s *SidecarEgressMap) Delete(key string) {
 	delete(s.cache, key)
 }
 
+// Map func returns a map of identity to namespace:SidecarEgress map
+// Iterating through the returned map is not implicitly thread safe,
+// use (s *SidecarEgressMap) Range() func instead.
 func (s *SidecarEgressMap) Map() map[string]map[string]SidecarEgress {
 	return s.cache
+}
+
+// Range is a thread safe iterator to iterate through the SidecarEgress map
+func (s *SidecarEgressMap) Range(fn func(k string, v map[string]SidecarEgress)) {
+	s.mutex.Lock()
+	for k, v := range s.cache {
+		fn(k, v)
+	}
+	s.mutex.Unlock()
 }
--- a/admiral/pkg/controller/common/types_test.go
+++ b/admiral/pkg/controller/common/types_test.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -200,10 +201,14 @@ func TestSidecarEgressGet(t *testing.T) {
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(3*time.Second))
 	defer cancel()
 
+	var wg sync.WaitGroup
+	wg.Add(2)
+	// Producer go routine
 	go func(ctx context.Context) {
 		for {
 			select {
 			case <-ctx.Done():
+				wg.Done()
 				return
 			default:
 				egressMap.Put("pkey1", string(uuid.NewUUID()), "fqdn", map[string]string{"pkey2": "pkey2"})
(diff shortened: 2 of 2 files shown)


admiral/pkg/controller/common/types.go
```diff
@@ -177,7 +177,10 @@ func (s *SidecarEgressMap) Put(identity string, namespace string, fqdn string, c
 }
 
 func (s *SidecarEgressMap) Get(key string) map[string]SidecarEgress {
-	return s.cache[key]
+	s.mutex.Lock()
```
ccfishk, line 180:
```suggestion
	s.mutex.Lock()
	defer s.mutex.Unlock()
```

admiral/pkg/controller/common/types.go
```diff
@@ -186,6 +189,18 @@ func (s *SidecarEgressMap) Delete(key string) {
...
+// Map func returns a map of identity to namespace:SidecarEgress map
+// Iterating through the returned map is not implicitly thread safe,
+// use (s *SidecarEgressMap) Range() func instead.
 func (s *SidecarEgressMap) Map() map[string]map[string]SidecarEgress {
 	return s.cache
 }
+
+// Range is a thread safe iterator to iterate through the SidecarEgress map
+func (s *SidecarEgressMap) Range(fn func(k string, v map[string]SidecarEgress)) {
+	s.mutex.Lock()
```
ccfishk, line 201: defer s.mutex.Unlock

admiral/pkg/controller/common/types_test.go
```diff
@@ -191,3 +191,50 @@ func TestMapRange(t *testing.T) {
...
+
+func TestSidecarEgressGet(t *testing.T) {
+
+	egressMap := NewSidecarEgressMap()
+	egressMap.Put("pkey1", "pkey2", "fqdn", map[string]string{"pkey2": "pkey2"})
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(3*time.Second))
+	defer cancel()
+
+	go func(ctx context.Context) {
```
nirvanagit, line 203: Adding a few comments in the relevant sections of the code will be useful.

admiral/pkg/controller/common/types.go
```diff
@@ -186,6 +189,18 @@ func (s *SidecarEgressMap) Delete(key string) {
...
 
+// Map func returns a map of identity to namespace:SidecarEgress map
+// Iterating through the returned map is not implicitly thread safe,
+// use (s *SidecarEgressMap) Range() func instead.
 func (s *SidecarEgressMap) Map() map[string]map[string]SidecarEgress {
 	return s.cache
 }
+
+// Range is a thread safe iterator to iterate through the SidecarEgress map
+func (s *SidecarEgressMap) Range(fn func(k string, v map[string]SidecarEgress)) {
```
nirvanagit, line 200: Where will this be used?

nirvanagit, line 200: Nice!

shriramsharma: hey @nirvanagit , I've added this func to provide a thread-safe way to iterate through the `SidecarEgress` map. Basically discouraging using the `.Map()` func to get access to the underlying map and then performing a `for range` on it. I've added comments to make it clear.

admiral/pkg/controller/common/types_test.go
shriramsharma: @nirvanagit , I wrote this test to reproduce the concurrent map access failure. Does it need a comment?

ccfishk, line 203:
```suggestion
       wg := synced.WaitGroup()
       wg.Add(1)
	go func(ctx context.Context) {
	defer wg.Done()
```
Understand it is test code, but to avoid possible memory leak, waitgroup can help in this case.

ccfishk (approved):
Kindly address comments

admiral/pkg/controller/common/types_test.go
shriramsharma: Thanks @ccfishk . I'm sorry I'm trying to understand where could this lead to a memory leak. As per my understanding, as the context's deadline expires, `ctx.Done()`  would get called and the `go routine` in that case should return. Please correct me if I'm wrong.

ccfishk: @shriramsharma  Sorry for the late response. 
I mean goroutine leak, which is memory leak, coz goroutine is not garbage cleaned by runtime object, only if the parent process specifically terminate the  goroutine before it exit. 
One example is that parent process stops for error while goroutine is left as it is. The clean way includes
 - signal (like channle close make(chan interface{})) between parent and the routine,
 -  waitgroup  synchronize

shriramsharma: @ccfishk , I think I understand where you're getting at. Thanks. The issue is with the unpredictable order in which the `Done()` method will be called. There could be a scenario where the parent go routine exits prior to the child. 
I added `Waitgroup` as you suggested.

admiral/pkg/controller/common/types.go
nirvanagit: Is this already being used ?

admiral/pkg/controller/common/types_test.go
nirvanagit: Just adding comments on the producer and consumer side stating what is doing what, will enhance readability for everyone, even if they are not super familiar with channels.

nirvanagit (approved):

admiral/pkg/controller/common/types.go
shriramsharma: No it's not being used right now

PR #225, opened 2022-06-06
concurrent map issue

shriramsharma:
fixes 224
Added mutex to synchronize access to `Get` func.

### DIFF
--- a/admiral/pkg/controller/admiral/service.go
+++ b/admiral/pkg/controller/admiral/service.go
@@ -2,17 +2,19 @@ package admiral
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/istio-ecosystem/admiral/admiral/pkg/controller/common"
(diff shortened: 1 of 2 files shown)


vinay-g (approved):

admiral/pkg/controller/admiral/service.go
```diff
@@ -69,6 +71,8 @@ func (s *serviceCache) getKey(service *k8sV1.Service) string {
 }
 
 func (s *serviceCache) Get(key string) *ServiceClusterEntry {
+	defer s.mutex.Unlock()
```
nirvanagit, line 74: nitpick: should we reverse them? Have `s.mutext.Lock()` and then `defer` ?

ccfishk, line 74:
```suggestion
         s.mutex.Lock()
	defer s.mutex.Unlock()
```

ccfishk (approved):

Issue #234, opened 2022-06-24
[BUG]: Deleting a GTP would result in deleting the only endpoint of a service


shriramsharma:
**Describe the bug**
With this PR https://github.com/istio-ecosystem/admiral/pull/223 we have an issue where a delete GTP event could cause the only service entry endpoint to be deleted since we're passing down the same admiral event down to the `modifyServiceEntryForNewServiceOrPod` func.

PR #235, opened 2022-06-24
force GTPs to update only

shriramsharma:
Fixes: #234
force GTPs to update only

Signed-off-by: Shriram Sharma <shriram_sharma@intuit.com>


aattuluri (approved):
lgtm

PR #240, opened 2022-07-05
Fix event processing during bootup

aattuluri:


admiral/pkg/clusters/serviceentry.go
```diff
@@ -58,6 +58,10 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
 
 	defer util.LogElapsedTime("modifyServiceEntryForNewServiceOrPod", sourceIdentity, env, "")()
 
+	if IsCacheWarmupTime(remoteRegistry) {
+		log.Debugf(LogFormat, event, env, sourceIdentity, "", "Processing skipped during cache warm up state")
```
nirvanagit, line 62: should we make it `log.Infof`

nirvanagit (approved):

PR #243, opened 2022-07-08
Routing policy addition

vinay-g:
Rework from - https://github.com/istio-ecosystem/admiral/pull/229


admiral/pkg/clusters/types.go
```diff
@@ -127,11 +130,185 @@ func (g *globalTrafficCache) Put(gtp *v1.GlobalTrafficPolicy) error {
...
+
+func (r RoutingPolicyHandler) Deleted(obj *v1.RoutingPolicy) {
+	log.Info("deleted envoy filter for routing policy - ",obj.Name)
+	dependents, missingIdentityLabel := getDependents(obj, r)
+	if !missingIdentityLabel {
+		r.deleteEnvoyFilters(dependents, obj, admiral.Delete)
+	}
+}
+
+func (r RoutingPolicyHandler) deleteEnvoyFilters(dependents map[string]string, obj *v1.RoutingPolicy, eventType admiral.EventType) {
```
shriramsharma, line 291: @vinay-g is it possible to return the `error` to the caller instead of logging it in this func. I feel the logging should happen at the caller.

admiral/pkg/clusters/types.go
```diff
@@ -127,11 +130,185 @@ func (g *globalTrafficCache) Put(gtp *v1.GlobalTrafficPolicy) error {
...
+		log.Warnf(LogErrFormat, "add", "RoutingPolicy", obj.Name, r.ClusterID, err)
+		return nil, true
+	}
+
+	dependents := r.RemoteRegistry.AdmiralCache.IdentityDependencyCache.Get(sourceIdentity).Copy()
+	return dependents, false
+}
+
+func (r RoutingPolicyHandler) Deleted(obj *v1.RoutingPolicy) {
+	log.Info("deleted envoy filter for routing policy - ",obj.Name)
```
shriramsharma, line 284: This log statement can be wrapped inside an `if err != nil` once `deleteEnvoyFilters` is modified to return an error.

admiral/pkg/clusters/types.go
```diff
@@ -127,11 +130,185 @@ func (g *globalTrafficCache) Put(gtp *v1.GlobalTrafficPolicy) error {
...
+	}
+	dependents, missingIdentityLabel := getDependents(obj, r)
+	if missingIdentityLabel {
+		return
+	}
+	r.processroutingPolicy(dependents, obj, admiral.Update)
+	log.Info("updated routing policy")
+}
+
+func getDependents(obj *v1.RoutingPolicy, r RoutingPolicyHandler) (map[string]string, bool) {
```
shriramsharma, line 271: just a nitpick, the `bool` returned by the func is a little confusing, at least to me. Can we add a comment to this which explains when/why would the func return `true` vs `false`?

admiral/pkg/clusters/types.go
```diff
@@ -127,11 +130,185 @@ func (g *globalTrafficCache) Put(gtp *v1.GlobalTrafficPolicy) error {
...
+	}
+	dependents, done := getDependents(obj, r)
+	if done {
+		return
+	}
+	r.processroutingPolicy(dependents, obj, admiral.Add)
+	log.Info("finished processing routing policy")
+}
+
+func (r RoutingPolicyHandler) processroutingPolicy(dependents map[string]string, routingPolicy *v1.RoutingPolicy, eventType admiral.EventType ) {
```
shriramsharma, line 236: same thing here, can we return the `error` from this func rather than logging it here?
That way, line #268 can be wrapped in an `if err` block.

admiral/pkg/clusters/registry_test.go
```diff
@@ -187,10 +188,22 @@ func createMockRemoteController(f func(interface{})) (*RemoteController, error)
 	}
 	stop := make(chan struct{})
 	d, e := admiral.NewDeploymentController("", stop, &test.MockDeploymentHandler{}, &config, time.Second*time.Duration(300))
+	if e != nil {
```
shriramsharma, line 191: very nitpicky, this can be ignored, but is it possible to change `e` to `err`?

admiral/pkg/clusters/types.go
vinay-g: done

admiral/pkg/clusters/types.go
vinay-g: removed it

admiral/pkg/clusters/types.go
vinay-g: done

admiral/pkg/clusters/registry_test.go
vinay-g: done

admiral/pkg/clusters/envoyfilter.go
```diff
@@ -0,0 +1,165 @@
...
+	"istio.io/api/networking/v1alpha3"
+	networking "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
+)
+
+var (
+	getSha1 = common.GetSha1
+)
+const wasmPath = "/etc/istio/extensions/dynamicrouter.wasm"
```
aattuluri, line 19: Can this be moved to the RoutingPolicy as well?

admiral/pkg/clusters/registry.go
```diff
@@ -188,6 +194,54 @@ func (r *RemoteRegistry) createCacheController(clientConfig *rest.Config, cluste
...
 	}
+	
+	log.Infof("starting Routing Policies controller for custerID: %v", clusterID)
+	rc.RoutingPolicyController, err = admiral.NewRoutingPoliciesController(stop, &RoutingPolicyHandler{RemoteRegistry: r, ClusterID: clusterID}, clientConfig, 1 * time.Minute)
+
+	if err != nil {
+		return fmt.Errorf("error with virtualServiceController init: %v", err)
+	}
+
+	log.Infof("starting node controller clusterID: %v", clusterID)
```
aattuluri, line 205: Why are all these controllers being restarted?

admiral/pkg/controller/common/common.go
```diff
@@ -38,6 +42,7 @@ const (
 	RolloutPodHashLabel           = "rollouts-pod-template-hash"
 	RolloutActiveServiceSuffix	  = "active-service"
 	RolloutStableServiceSuffix	  = "stable-service"
+	EnvoyFilterSuffix             = "1.13"
```
aattuluri, line 45: Why hardcode this, we are getting it from the input params right?

install/sample/sample_rp.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  plugin: greeting
+  hosts:
+    - qal.greeting.mesh
+  config:
+    cachePrefix: "prefix-v1-a"
+    cacheTtlSec: "1440"
+    dispatcherUrl: "http://test.does.not.matter.mesh/"
+    env: "qal"
+    algorithmName: "QBODispatcher"
```
aattuluri, line 18:
```suggestion
    algorithmName: "ShardedRouting"
```

aattuluri (changes requested):
Overall looks ok, but added some comments to refactor and/or changes

install/sample/sample_rp.yaml
vinay-g: done

admiral/pkg/controller/common/common.go
vinay-g: removed

admiral/pkg/clusters/envoyfilter.go
vinay-g: moved, but it will be there in user config so we need to make them aware

admiral/crd/routingPolicy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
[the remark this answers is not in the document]
vinay-g: done

admiral/cmd/admiral/cmd/root.go
```diff
@@ -131,7 +131,10 @@ func GetRootCmd(args []string) *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GatewayApp, "gateway_app", "istio-ingressgateway",
 		"The the value of the `app` label to use to match and find the service that represents the ingress for cross cluster traffic (AUTO_PASSTHROUGH mode)")
 	rootCmd.PersistentFlags().BoolVar(&params.MetricsEnabled, "metrics", true, "Enable prometheus metrics collections")
-
+	rootCmd.PersistentFlags().StringVar(&params.EnvoyFilterVersion, "envoy_filter_version", "^1\\.10|3.*",
+		"The value of envoy filter version is used to match the proxy version for envoy filter created by routing policy")
+	rootCmd.PersistentFlags().StringVar(&params.EnvoyFilterAdditionalConfig, "envoy_filter_additional_config", "dynamicRouterHost: internal.intuit.services.mesh.dynamicroutingservice.mesh",
+		"The value of envoy filter is to add additional config to the filter config section")
```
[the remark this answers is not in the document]
vinay-g: The dynamic routing service endpoint would be same, the internal dispatcher URL can be different. This is specific to our service so we can code it into the deployment spec.

admiral/cmd/admiral/cmd/root.go
```diff
@@ -131,7 +131,10 @@ func GetRootCmd(args []string) *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.GatewayApp, "gateway_app", "istio-ingressgateway",
 		"The the value of the `app` label to use to match and find the service that represents the ingress for cross cluster traffic (AUTO_PASSTHROUGH mode)")
 	rootCmd.PersistentFlags().BoolVar(&params.MetricsEnabled, "metrics", true, "Enable prometheus metrics collections")
-
+	rootCmd.PersistentFlags().StringVar(&params.EnvoyFilterVersion, "envoy_filter_version", "^1\\.10|3.*",
```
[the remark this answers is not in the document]
vinay-g: Will add this in next commit, looking into integration tests

admiral/pkg/clusters/registry.go
vinay-g: will check

install/admiral/base/roles.yaml
```diff
@@ -33,3 +33,15 @@ rules:
...
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: admiral-sync-write-envoyfilters
+  namespace: admiral
+rules:
+  - apiGroups: ["networking.istio.io"]
+    resources: ['envoyfilters']
+    verbs: ["create", "update", "delete"]
```
nirvanagit, line 47: do we need patch?

admiral/pkg/controller/common/common.go
```diff
@@ -38,6 +42,7 @@ const (
 	RolloutPodHashLabel           = "rollouts-pod-template-hash"
 	RolloutActiveServiceSuffix	  = "active-service"
 	RolloutStableServiceSuffix	  = "stable-service"
+	WASMPath					  = "wasmPath"
```
nirvanagit, line 45: nitpick: indentation

admiral/pkg/clusters/types.go
```diff
@@ -190,11 +202,206 @@ func (g *globalTrafficCache) Put(gtp *v1.GlobalTrafficPolicy) error {
...
+	ClusterID	   string
+}
+
+type routingPolicyCache struct {
+	// map of routing policies key=environment.identity, value: RoutingPolicy object
+	// only one routing policy per identity + env is allowed
+	identityCache map[string]*v1.RoutingPolicy
+
+	// map of dependent identity + env -> [] routingPolicy. There can be potentially multiple routingPolicies that can apply to a client.
+	// dependentRpCache map[string][]*v1.RoutingPolicy
```
nirvanagit, line 221: nitpick: remove dead code?

admiral/pkg/clusters/envoyfilter.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	admiralCache.RoutingPolicyFilterCache.Put(workloadIdentityKey+common.GetRoutingPolicyEnv(routingPolicy), rc.ClusterID, envoyFilterName)
+	var filter *networking.EnvoyFilter
+	//get the envoyfilter if it exists. If it exists, update it. Otherwise create it.
+	if eventType == admiral.Add || eventType == admiral.Update {
+		// We query the API server instead of getting it from cache because there could be potential condition where the filter exists in the cache but not on the cluster.
+		filter, err = rc.RoutingPolicyController.IstioClient.NetworkingV1alpha3().EnvoyFilters(common.NamespaceIstioSystem).Get(envoyFilterName, metaV1.GetOptions{})
+		if err != nil {
+			log.Infof("msg=%s filtername=%s clustername=%s", "creating the envoy filter", envoyFilterName, rc.ClusterID)
```
nirvanagit, line 55: should we use https://github.com/kubernetes/apimachinery/blob/117bd9b56ec3fe6b4891d86097971e49bf9f68f4/pkg/api/errors/errors.go#L527 to ensure that it is not there? I think your intention is to call Create irrespective, as best effort. Which is good too.

vinay-g: yup that was the idea let me test a little more in pre-prod to check if I can catch issues with always create

PR #245, opened 2022-07-14
Fix concurrent map access and remove unwanted code

aattuluri:
Fix concurrent map access on multiple controllers. 
Remove unwanted code to create SEs based on user created destination rules.
Add `-race` to detect races in tests.


admiral/pkg/clusters/registry_test.go
```diff
@@ -48,7 +48,7 @@ func TestDeleteCacheControllerThatDoesntExist(t *testing.T) {
 
 	w := RemoteRegistry{
 		RemoteControllers: make(map[string]*RemoteController),
-		StartTime: time.Now(),
+		StartTime:         time.Now(),
```
shriramsharma, line 51: nitpick, formatting issues.

admiral/pkg/clusters/registry_test.go
```diff
@@ -62,7 +62,7 @@ func TestDeleteCacheController(t *testing.T) {
...
```
shriramsharma, line 65: nitpick, formatting issues.

shriramsharma (approved):

admiral/pkg/clusters/registry_test.go
aattuluri: Seems to be the right formatting now. Was messed up before.

Issue #228, opened 2022-06-15
[BUG] Issues with multi-cluster set up


aattuluri:
**Describe the bug**
The multi-cluster set up docs aren't clear and there are at least a couple of issues when users tried the set up

alexdotsh:
Hi, is there any update on this to support k8s v1.22?

Issue #218, opened 2022-05-13
[BUG] Admiral install fails on k8s v1.22


levaitamas:
I was following [the install guide in Examples](https://github.com/istio-ecosystem/admiral/blob/master/docs/Examples.md#install) using Istio v1.13.3 and k8s v1.22.9 (kubectl v1.23.6).

At Step 4, the process fails due to not matching "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1":

```
$ kubectl apply --context=$REMOTE_CLUSTER -f $ADMIRAL_HOME/yaml/remotecluster.yaml
namespace/admiral-sync unchanged
serviceaccount/admiral unchanged
role.rbac.authorization.k8s.io/admiral-sync-write unchanged
clusterrole.rbac.authorization.k8s.io/admiral-sidecar-write unchanged
clusterrole.rbac.authorization.k8s.io/admiral-sync-read unchanged
clusterrole.rbac.authorization.k8s.io/istio-admin unchanged
clusterrole.rbac.authorization.k8s.io/istio-viewers unchanged
rolebinding.rbac.authorization.k8s.io/admiral-sync-write-binding unchanged
clusterrolebinding.rbac.authorization.k8s.io/admiral-sidecar-write-binding unchanged
clusterrolebinding.rbac.authorization.k8s.io/admiral-sync-read-binding unchanged
error: unable to recognize "/home/levait/istio-mc/admiral-install-v1.2/yaml/remotecluster.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

Turns out this is the expected behavior because of the API removals in k8s v1.22: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes

PR #232, opened 2022-06-22
Upgraded the dependent go modules

shriramsharma:
Fixes #216 
Upgraded the dependent go modules

Signed-off-by: Shriram Sharma <shriram_sharma@intuit.com>

PR #247, opened 2022-07-21
Update CRD to use apiextensions.k8s.io/v1

alexdotsh:
As of K8s 1.22 the apiextensions.k8s.io/v1beta1 is deprecated and longer used, which causing issues to install on K8s running on version 1.22

This is PR is to address and update to use the newer `apiextensions.k8s.io/v1`

The apiextensions.k8s.io/v1 API version, available since v1.16.

#228
#218


aattuluri (approved):

aattuluri:
@alexdotsh Thanks for the PR. This looks good, would you mind auditing other resources under the admiral/install directory for other resources that would have been deprecated in k8s 1.22. 
FYI, there is another [PR](https://github.com/istio-ecosystem/admiral/pull/232) to update the k8s client in Admiral which `might` also be required for Admiral to work for k8s 1.22 CC @shriramsharma

alexdotsh:
@aattuluri those were the only ones I manage to find that was deprecated. I think the rest are fine

shriramsharma:
@alexdotsh , can you please resolve the CI failures? Thanks

PR #249, opened 2022-07-25
Support for GlobalTrafficPolicy priority processing

vrushalijoshi:
Adding support to handle GlobalTrafficPolicy with priority label
1. Added label for resource priority processing (currently used for only processing GlobalTrafficPolicies)
2. Code updates to handle priority labeled GlobalTrafficPolicy resource, and pick the most recently created one if there are multiple priority labeled GTPs.


admiral/pkg/clusters/serviceentry.go
```diff
@@ -242,30 +242,44 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
...
+			}
+			mostRecentGtp = gtpsOrdered[0]
+		} else {
+			if len(priorityGtpsOrdered) > 1 {
+				log.Infof("More than one GTP with priority label found for identity=%s in env=%s.", identity, env)
+				sortFetchedGtps(priorityGtpsOrdered, identity, env)
+			}
+			mostRecentGtp = priorityGtpsOrdered[0]
+		}
+	}
```
aattuluri, line 282: Can't we build this into sorting function itself? Sort by priority label present followed by date created?

aattuluri (changes requested):
Please take a look at the comments.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -242,30 +242,44 @@ func modifyServiceEntryForNewServiceOrPod(event admiral.EventType, env string, s
...
-		sort.Slice(gtpsOrdered, func(i, j int) bool {
-			iTime := gtpsOrdered[i].CreationTimestamp
-			jTime := gtpsOrdered[j].CreationTimestamp
-			log.Debugf("GTP sorting identity=%s env=%s name1=%s creationTime1=%v name2=%s creationTime2=%v", identity, env, gtpsOrdered[i].Name, iTime, gtpsOrdered[j].Name, jTime)
-			return iTime.After(jTime.Time)
-		})
-	}
+	} else {
+		for _, matchedGtp := range gtpsOrdered {
+			if matchedGtp.ObjectMeta.Labels[common.GetAdmiralParams().LabelSet.ResourcePriorityKey] == "true" {
```
shriramsharma, line 264: @vrushalijoshi , just a thought. This would mean that any GTP created by "anyone" with `admiral.io/resource-priority: true` would be treated as a higher priority. 
Do you think it would make sense to have a value that could be passed as a param when the admiral starts and we only look for that value in the label?

vrushalijoshi: Added fix for priority to value instead of true/false

admiral/pkg/clusters/serviceentry.go
vrushalijoshi: Updated the sorting function

admiral/pkg/clusters/serviceentry.go
```diff
@@ -276,6 +290,15 @@ func updateGlobalGtpCache(cache *AdmiralCache, identity, env string, gtps map[st
 	}
 }
 
+func sortFetchedGtps(gtpsToOrder []*v1.GlobalTrafficPolicy, identity string, env string) {
```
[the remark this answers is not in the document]
vrushalijoshi: Updated the method name

admiral/cmd/admiral/cmd/root.go
```diff
@@ -115,6 +115,8 @@ func GetRootCmd(args []string) *cobra.Command {
 		"The label value, on a namespace or service, which tells Istio to perform sidecar injection")
 	rootCmd.PersistentFlags().StringVar(&params.LabelSet.AdmiralIgnoreLabel, "admiral_ignore_label", "admiral-ignore",
 		"The label value, on a namespace, which tells Istio to perform sidecar injection")
+	rootCmd.PersistentFlags().StringVar(&params.LabelSet.ResourcePriorityKey, "resource_priority_key", "admiral.io/resource-priority",
```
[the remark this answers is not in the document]
vrushalijoshi: updated to priority label

admiral/pkg/clusters/registry_test.go
```diff
@@ -40,6 +40,7 @@ func init() {
 
 	p.LabelSet.WorkloadIdentityKey = "identity"
 	p.LabelSet.GlobalTrafficDeploymentLabel = "identity"
+	p.LabelSet.ResourcePriorityKey = "admiral.io/resource-priority"
```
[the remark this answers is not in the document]
vrushalijoshi: updated to priority label

aattuluri (approved):
lgtm

PR #253, opened 2022-08-09
fix: Rollout canary when traffic routing not defined

itsLucario:
When canary.trafficRouting is not defined for Argo rollout canary strategy.

**Expected:**
The service match should return whichever service has a mesh port defined for creating service entry.

**Found:**
Admiral by default tries to match with the stable K8s service defined within rollout.stableService and if mesh port is not found in that K8s service, the Service Entry creation fails.

**Fix:**
When `canary.trafficRouting` is not defined for Argo rollout canary strategy. Scenario 1 from below should be expected.

**Overall Expected:**
Scenario 1 - canary strategy defined and stable service is non empty
Pick stable service if present otherwise first service that matches
Scenario 2 -  canary strategy defined with istio traffic management
Pick stable service/canary services defined in canary strategy
Senario 3 - No stable service is defined under canary strategy then pick a first matching service

### DIFF
--- a/admiral/pkg/clusters/handler.go
+++ b/admiral/pkg/clusters/handler.go
@@ -726,6 +726,8 @@ func getServiceForRollout(rc *RemoteController, rollout *argo.Rollout) map[strin
 	var blueGreenActiveService string
 	var blueGreenPreviewService string
 
+	var matchedServices = make(map[string]*WeightedService)
+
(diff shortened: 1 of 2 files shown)


aattuluri (approved):
lgtm overall, will address optimizations in later PR.

PR #254, opened 2022-08-26
log error, instead of returning when vs deletion fails

nirvanagit:
## Goal
When a replicated VS is deleted in the source cluster, it gets deleted in clusters where it is replicated in. During this process, if the deletion fails due to some reason, it skips deleting the vs from the remaining clusters. 

This PR changes that behavior, and makes it consistent with deletion workflow for DestinationRules, and ServiceEntries, where it logs that there was an error and continues deleting the resource in other clusters.


aattuluri (changes requested):
The change looks good. Should we add a unit with mock controller one to deleting an existing resource and another to delete a non existing one?

nirvanagit:
> The change looks good. Should we add a unit with mock controller one to deleting an existing resource and another to delete a non existing one?

done.

aattuluri (approved):
lgtm

PR #255, opened 2022-08-30
Stop creation of service entires with dummy endpoints

vrushalijoshi:
Updates to avoid creation of service entry with dummy endpoints. Include updates to delete existing service entry with only dummy endpoints.

### DIFF
--- a/admiral/pkg/clusters/handler.go
+++ b/admiral/pkg/clusters/handler.go
@@ -504,6 +504,21 @@ func addUpdateVirtualService(ctx context.Context, obj *v1alpha3.VirtualService, 
 	}
 }
 
+func validateServiceEntryEndpoints(endpoints []*v1alpha32.WorkloadEntry) (bool, []*v1alpha32.WorkloadEntry) {
+	var areEndpointsValid = true
+	validEndpoints := make([]*v1alpha32.WorkloadEntry, 0)
+
+	for _, endpoint := range endpoints {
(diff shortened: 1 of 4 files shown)


admiral/pkg/clusters/handler.go
```diff
@@ -515,13 +530,23 @@ func addUpdateServiceEntry(ctx context.Context, obj *v1alpha3.ServiceEntry, exis
...
+
 	if exist == nil || exist.Spec.Hosts == nil {
-		obj.Namespace = namespace
-		obj.ResourceVersion = ""
-		_, err = rc.ServiceEntryController.IstioClient.NetworkingV1alpha3().ServiceEntries(namespace).Create(ctx, obj, v12.CreateOptions{})
 		op = "Add"
-		log.Infof(LogFormat+" SE=%s", op, "ServiceEntry", obj.Name, rc.ClusterID, "New SE", obj.Spec.String())
-	} else {
+		//se will be created if endpoints are valid, in case they are not valid se will be created with just valid endpoints
+		if areEndpointsValid || (!areEndpointsValid && len(validEndpoints) > 0) {
```
aattuluri, line 540: It would be simpler if we can filter them inline during validation. Log any filtered endpoints in the method itself.

aattuluri (changes requested):
The functionality looks good. I think it can be simplified.

aattuluri (approved):
lgtm

PR #256, opened 2022-08-30
check for nil pointer before invoking rollout controller

nirvanagit:
During startup time, the virtual service controller and the rollout controller get initialized successfully, but the remote registry cache (which stores all the controllers) is initialized only after all the controllers have been initialized, as seen in https://github.com/istio-ecosystem/admiral/blob/bdfcbae2337a6314dbd8c4a2fc9b23db3509abcb/admiral/pkg/clusters/registry.go#L237-L238

Meanwhile the controllers are getting initialized, there might be a virtual service event which gets sent from the kubernetes API server. When handling a virtual service event, admiral checks for corresponding argorollout. But until this https://github.com/istio-ecosystem/admiral/blob/bdfcbae2337a6314dbd8c4a2fc9b23db3509abcb/admiral/pkg/clusters/registry.go#L237-L238 code is not called, the remote registry will not be cached, hence when this statement is executed: https://github.com/istio-ecosystem/admiral/blob/05fc3dbbba4757c03be28a52f67c7316402b32fc/admiral/pkg/clusters/handler.go#L390, it causes a nil pointer exception.

In order to prevent the nil pointer, we explicitly check if the remote registry cache has the controller for the given cluster, and if it exist, we continue. Else the sync will happen in the next full sync cycle.


aattuluri (commented):
Looks good. I am worried this this can happen for other events like DestinationRule etc when they are added to customer namespaces.
Should we implement this in the controllers one level up, do we have access to RemoteRegistry there?

shriramsharma (approved):

nirvanagit:
> Looks good. I am worried this this can happen for other events like DestinationRule etc when they are added to customer namespaces. Should we implement this in the controllers one level up, do we have access to RemoteRegistry there?

Totally missed this one, did you mean in Handlers of each type? If so, yes, each handler has remote registry as its property.

PR #260, opened 2022-09-19
Feature to exclude an asset via command line parameter

nirvanagit:

PR #261, opened 2022-09-19
Feature to exclude an identity via command line parameter #260 

nirvanagit:


shriramsharma (approved):

admiral/pkg/clusters/types.go
```diff
@@ -151,6 +154,15 @@ func (r *RemoteRegistry) shutdown() {
 	}
 }
 
+func isAnExcludedAsset(assetName string, excludedAssetList []string) bool {
```
aattuluri, line 157: Can you convert `excludedAssetList` into a map?
```suggestion
func isAnExcludedAsset(assetName string, excludedIdentityList []string) bool {
```

aattuluri (changes requested):

aattuluri (approved):
Lgtm

PR #265, opened 2022-11-07
fix: Handle VS hosts nil

itsLucario:
VS hosts field is an optional field in Istio's VirtualService CRD spec. It'll be nil when we want to create a delegate Virtual Service. 
The fix will do a `nil` check and if the VS hosts are nil, the admiral will replicate the VS resource as it is within all the clusters.


admiral/pkg/clusters/handler.go
```diff
@@ -418,49 +418,51 @@ func handleVirtualServiceEvent(
...
-		for _, dependentCluster := range dependentClusters {
-			rc := r.GetRemoteController(dependentCluster)
-			if clusterId != dependentCluster {
-				log.Infof(LogFormat, "Event", "VirtualService", obj.Name, clusterId, "Processing")
-				if event == common.Delete {
-					err := rc.VirtualServiceController.IstioClient.NetworkingV1alpha3().VirtualServices(syncNamespace).Delete(ctx, obj.Name, v12.DeleteOptions{})
-					if err != nil {
-						if k8sErrors.IsNotFound(err) {
-							log.Infof(LogFormat, "Delete", "VirtualService", obj.Name, clusterId, "Either VirtualService was already deleted, or it never existed")
+	if virtualService.Hosts != nil {
```
nirvanagit, line 421: Its better to check the length of the slice, because if an empty `[]string` is assigned to `virtualService.Hosts` then it will not be nil.

itsLucario: Ah, Right!!  I've updated

aattuluri (approved):
lgtm

PR #267, opened 2022-12-12
add interface for ignoring identity through command line

nirvanagit:
## Changes:
Add interface for endpoint generation suspension so that other implementations can be easily swapped with the current command line implementation.

### DIFF
--- a/admiral/pkg/clusters/endpointSuspender.go
+++ b/admiral/pkg/clusters/endpointSuspender.go
@@ -1,5 +1,0 @@
-package clusters
-
-type EndpointSuspender interface {
-	SuspendGeneration(identity string, environment string) bool
-}
--- a/admiral/pkg/clusters/endpointSuspender.go
+++ b/admiral/pkg/clusters/endpointSuspender.go
@@ -1,0 +1,5 @@
+package clusters
+
+type EndpointSuspender interface {
+	SuspendGeneration(identity string, environment string) bool
+}
--- a/admiral/pkg/clusters/endpointSuspension.go
+++ b/admiral/pkg/clusters/endpointSuspension.go
@@ -40,6 +40,14 @@ func NewDefaultEndpointSuspension(items []string) *endpointSuspension {
 	}}
 }
 
+func NewDummyEndpointSuspension() *endpointSuspension {
+	return &endpointSuspension{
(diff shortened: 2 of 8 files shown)


admiral/pkg/clusters/endpointSuspender.go
```diff
@@ -0,0 +1,5 @@
+package clusters
+
+type EndpointSuspender interface {
```
aattuluri, line 3: Refactor this to EndpointUpdateSuspender?

admiral/pkg/clusters/endpointSuspension.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (des *endpointSuspension) identityByEnvironment(identity, environment string) bool {
+	log.Printf("op=dynamicEndpointSuspension message=checking if identity %s in environment %s is in the suspension list",
+		identity, environment)
+	des.ignoredIdentityCache.RWLock.RLock()
+	defer des.ignoredIdentityCache.RWLock.RUnlock()
+	if des.ignoredIdentityCache.EnvironmentsByIdentity[identity] != nil {
+		identityEnvironments := des.ignoredIdentityCache.EnvironmentsByIdentity[identity]
+		if len(identityEnvironments) == 0 || (len(identityEnvironments) == 1 && identityEnvironments[0] == "") {
+			log.Printf("%s, identity: %s", alertMsgSuspensionForIdentityInAllEnvironments, identity)
```
aattuluri, line 79: `(len(identityEnvironments) == 1 && identityEnvironments[0] == "")` is this a special case or just a safety check?

aattuluri (changes requested):
lgtm, a small change requested and a question.

admiral/pkg/clusters/endpointSuspension.go
nirvanagit: Yes, this is the case when the entry is like:
```golang
identityByEnvironment: map[string][]string{
  "identity": []string{""},
}
```

admiral/pkg/clusters/endpointSuspender.go
nirvanagit: How about `ServiceEntryUpdateSuspender` ?

aattuluri (approved):
lgtm

PR #271, opened 2023-01-06
skipped updating istio resources which were not created by admiral

### LINKED ISSUE #270: Skip overriding SEs and DRs that were not created by admiral
**Describe the bug**
Currently, admiral overrides ServiceEntry and DestinationRule istio resources which admiral did not create. If a user-generated SE or DR follows the same naming semantics (`hostname-se/hostname-default-dr`) as admiral, then admiral overrides the resource.

**Steps To Reproduce**
1. Create an istio-enabled deployment with necessary admiral annotations.
2. Create a service for the above deployment
3. Verify if an SE is generated in the sync namespace.
4. Delete and recreate the SE with the same name and modify the endpoints.
5. Edit and make a minor change to the deployment. (Adding a new label/annotation)
6. Check the SE and you'll notice that the admiral would have overridden the SE.

**Expected behavior**
Since all SEs and DRs that admiral generates to have the following annotation `app.kubernetes.io/created-by: admiral`, any resource that does not have this annotation, should not be updated/deleted by admiral.

shriramsharma:
Fixes #270

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -36,8 +36,8 @@ type SeDrTuple struct {
 }
 
 const (
-	istioCustomResourceCreatedByAnnotationLabel = "app.kubernetes.io/created-by"
-	istioCustomResourceCreatedByAnnotationValue = "admiral"
+	resourceCreatedByAnnotationLabel = "app.kubernetes.io/created-by"
+	resourceCreatedByAnnotationValue = "admiral"
 )
 
 func createServiceEntryForDeployment(ctx context.Context, event admiral.EventType, rc *RemoteController, admiralCache *AdmiralCache,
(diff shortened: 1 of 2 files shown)


admiral/pkg/clusters/serviceentry.go
```diff
@@ -467,15 +472,40 @@ func AddServiceEntriesWithDr(ctx context.Context, rr *RemoteRegistry, sourceClus
 					log.Infof(LogFormat, "Get (error)", "old ServiceEntry", seDr.SeName, sourceCluster, err)
 					oldServiceEntry = nil
 				}
+
+				// check if the existing service entry was created outside of admiral
+				// if it was, then admiral will not take any action on this SE
+				skipSEUpdate := false
+				if oldServiceEntry != nil {
+					if !isGeneratedByAdmiral(oldServiceEntry.Annotations) {
```
aattuluri, line 480: Can be made a single if statement instead of nesting?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -484,29 +514,45 @@ func AddServiceEntriesWithDr(ctx context.Context, rr *RemoteRegistry, sourceClus
...
-					if newServiceEntry != nil {
-						newServiceEntry.Labels = map[string]string{common.GetWorkloadIdentifier(): fmt.Sprintf("%v", identityId)}
-						addUpdateServiceEntry(ctx, newServiceEntry, oldServiceEntry, syncNamespace, rc)
-						cache.SeClusterCache.Put(newServiceEntry.Spec.Hosts[0], rc.ClusterID, rc.ClusterID)
+					if !skipSEUpdate {
+						//nolint
+						newServiceEntry := createServiceEntrySkeletion(*seDr.ServiceEntry, seDr.SeName, syncNamespace)
+						if newServiceEntry != nil {
+							newServiceEntry.Labels = map[string]string{common.GetWorkloadIdentifier(): fmt.Sprintf("%v", identityId)}
+							addUpdateServiceEntry(ctx, newServiceEntry, oldServiceEntry, syncNamespace, rc)
```
aattuluri, line 531: Is it possible to move the logic to skip updating into `addUpdateServiceEntry` and `addUpdateDestinationRule`? That might make it cleaner.

aattuluri (changes requested):
The implementation looks good, couple of refactor suggestions.

admiral/pkg/clusters/serviceentry.go
shriramsharma: hey @aattuluri , I agree it will make it cleaner but I would need to do the same for the `deleteSE` and `deleteDR` funcs. Plus, in order to avoid adding it to the SEClusterCache, I added the logic here.

aattuluri: Got it, I am good then.

aattuluri (approved):
lgtm

PR #279, opened 2023-01-24
Add dependencyproxy model, crds, listers, informers

### LINKED ISSUE #278: [FEATURE] - Add an ability to provide a mesh service that acts as a proxy through which dependencies should be accessed
**Is your feature request related to a problem? Please describe.**
Need the ability to automatically forward requests to a mesh dependency through a proxy service. 

**Describe the solution you'd like**
Create a new CRD that contains the configuration of how a destination can be reached through a given proxy. The configuration also provides the ability to generate additional hostnames through which a service can be called.

Example CRD:
```
apiVersion: admiral.io/v1alpha1
kind: DependencyProxy
metadata:
  name: dependency-proxy-example
  namespace: admiral
  annotations:
     admiral.io/env: stage
spec:
  destination:
    identity: greeting
    dns_suffix: "xyz"
    dns_prefix: 
        - "test00.foo"
        - "test01.bar"
  proxy:
    identity: nginx-proxy
```

This generates the following ServiceEntry
```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: stage.nginx.global-se
spec:
  addresses:
  - 240.0.10.1
  endpoints:
  - address: nginx-gw-svc.default.svc.cluster.local
    locality: us-west-2
    ports:
      http: 80
  hosts:
  - test00.foo.stage.greeting.xyz
  - test01.bar.stage.greeting.xyz
```

shriramsharma:
fixes #278


admiral/pkg/apis/admiral/model/dependencyproxy.proto
```diff
@@ -0,0 +1,53 @@
...
+    repeated string dns_prefixes = 2;
+
+    // The DNS suffix that should be appended while
+    // constructing the endpoint of the destination service.
+    string dns_suffix = 3;
+}
+
+message Proxy {
+    // Identifier of the proxy's workload
+    string identity = 3;
```
aattuluri, line 52: Just to confirm, admiral.io/env label on the dependency proxy would be used in addition to this identity to construct the proxy endpoint?

admiral/crd/dependencyproxies.yaml
```diff
@@ -0,0 +1,75 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: dependencyproxies.admiral.io
+spec:
+  group: admiral.io
+  names:
+    kind: DependencyProxy
+    listKind: DependencyProxyList
+    plural: dependencyproxies
+    singular: dependencyproxy
```
aattuluri, line 11: Can we add a shortName for the CRD, something like `dp`?

aattuluri (changes requested):

admiral/pkg/apis/admiral/model/dependencyproxy.proto
shriramsharma: yes, that is correct. Let me know if you want to make it more explicit to avoid confusion.

admiral/crd/dependencyproxies.yaml
shriramsharma: yes, will do that

admiral/pkg/apis/admiral/model/dependencyproxy.proto
aattuluri: I think we can add an example in the docs with a sample dependencyproxy and what Admiral renders

admiral/pkg/apis/admiral/model/dependencyproxy.proto
```diff
@@ -0,0 +1,72 @@
...
+// VirtualService object
+//```yaml
+//  apiVersion: networking.istio.io/v1alpha3
+//  kind: VirtualService
+//  metadata:
+//      name: httpbin-vs
+//  spec:
+//      hosts:
+//       - test0.stage.greeting.xyz
+//       - test1.stage.greeting.xyz
```
aattuluri, line 36: I am thinking if these should be test0.greeting.xyz and test1.greeting.xyz because if some one has e2e1 and e2e2, the endpoints will look like e2e1.e2e.greeting.xyz and e2e2.e2e.greeting.xyz. Also this will make the migration later (from gateway to mesh seamless when that happens)

shriramsharma: hey @aattuluri , I'm sorry this example I think is a little confusing. `stage` here is the `admiral.io/env` on the deployment/rollout of the destination service(greeting). So I was thinking if we'll prepend the prefixes defined above to the generated endpoint of the service. `<env>.<identity>.<suffix>

So we'll be generating the regular endpoint without the prefix as well as shown here https://github.com/istio-ecosystem/admiral/blob/02fd5ef60ab3e3996addb37f1408e9a67aa9a5ec/admiral/pkg/apis/admiral/model/dependencyproxy.proto#L37
This should make the migration seamless. Please correct me if I'm wrong.

aattuluri: Makes sense.

aattuluri (approved):
lgtm

admiral/pkg/apis/admiral/v1/type.go
```diff
@@ -81,3 +81,31 @@ type RoutingPolicyList struct {
...
+type DependencyProxy struct {
+	meta_v1.TypeMeta   `json:",inline"`
+	meta_v1.ObjectMeta `json:"metadata"`
+	Spec               model.DependencyProxy `json:"spec"`
+	Status             DependencyProxyStatus `json:"status"`
+}
+
+// DependencyProxyStatus is the status for a DependencyProxy resource
+type DependencyProxyStatus struct {
+	ClusterSynced int32  `json:"clustersSynced"`
```
nirvanagit, line 100: What is the logic to update this?

admiral/pkg/controller/admiral/dependencyproxy.go
```diff
@@ -0,0 +1,113 @@
...
+	v1 "github.com/istio-ecosystem/admiral/admiral/pkg/apis/admiral/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	clientset "github.com/istio-ecosystem/admiral/admiral/pkg/client/clientset/versioned"
+	informerV1 "github.com/istio-ecosystem/admiral/admiral/pkg/client/informers/externalversions/admiral/v1"
+)
+
+// DepProxyHandler interface contains the methods that are required
+type DepProxyHandler interface {
```
nirvanagit, line 18: Dependency?

admiral/pkg/controller/admiral/dependencyproxy.go
```diff
@@ -0,0 +1,113 @@
...
+
+type DependencyProxyController struct {
+	K8sClient       kubernetes.Interface
+	DepCrdClient    clientset.Interface
+	DepProxyHandler DepProxyHandler
+	Cache           *depProxyCache
+	informer        cache.SharedIndexInformer
+}
+
+type depProxyCache struct {
```
nirvanagit, line 32: dependencyProxyCache?

admiral/pkg/controller/admiral/dependencyproxy.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (d *depProxyCache) Delete(dep *v1.DependencyProxy) {
+	defer d.mutex.Unlock()
+	d.mutex.Lock()
+	delete(d.cache, d.getKey(dep))
+}
+
+func NewDependencyProxyController(stopCh <-chan struct{}, handler DepProxyHandler, configPath string, namespace string, resyncPeriod time.Duration) (*DependencyProxyController, error) {
+
+	depProxyController := DependencyProxyController{}
```
nirvanagit, line 63: may be just controller?

admiral/pkg/controller/admiral/dependencyproxy.go
```diff
@@ -0,0 +1,113 @@
...
+
+	depProxyController.Cache = &depProxyCache
+	var err error
+
+	depProxyController.K8sClient, err = K8sClientFromPath(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create dependency controller k8s client: %v", err)
+	}
+
+	depProxyController.DepCrdClient, err = AdmiralCrdClientFromPath(configPath)
```
nirvanagit, line 78: Should we call it `admiralCRDClient`? I see its called DepCrdClient even for dependency.

admiral/pkg/controller/admiral/dependencyproxy.go
nirvanagit: Is the struct which implements this already in this PR?

admiral/pkg/apis/admiral/v1/type.go
shriramsharma: @nirvanagit , since this was in the other types, so I just kept it as is.
@aattuluri , do you know if we need this or if this can be removed?

aattuluri: @shriramsharma We should remove this given its doesnt make sense for this CRD.

nirvanagit (approved):

PR #281, opened 2023-02-06
Adding labels to admiral generated service entries

vrushalijoshi:
Update to SE modification flow, in order to add env, prefix and associated GTP details to SE labels.

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -529,11 +529,14 @@ func AddServiceEntriesWithDr(ctx context.Context, rr *RemoteRegistry, sourceClus
 								common.GetWorkloadIdentifier(): fmt.Sprintf("%v", identityId),
 								common.GetEnvKey():             fmt.Sprintf("%v", env),
 							}
+							if newServiceEntry.Annotations == nil {
+								newServiceEntry.Annotations = map[string]string{}
+							}
 							if seDr.SeDnsPrefix != "" && seDr.SeDnsPrefix != common.Default {
-								newServiceEntry.Labels["dnsPrefix"] = seDr.SeDnsPrefix
+								newServiceEntry.Annotations["dns-prefix"] = seDr.SeDnsPrefix
 							}
 							if seDr.SeDrGlobalTrafficPolicyName != "" {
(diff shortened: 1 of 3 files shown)


admiral/pkg/clusters/serviceentry.go
```diff
@@ -523,7 +525,16 @@ func AddServiceEntriesWithDr(ctx context.Context, rr *RemoteRegistry, sourceClus
...
 						newServiceEntry := createServiceEntrySkeletion(*seDr.ServiceEntry, seDr.SeName, syncNamespace)
 						if newServiceEntry != nil {
-							newServiceEntry.Labels = map[string]string{common.GetWorkloadIdentifier(): fmt.Sprintf("%v", identityId)}
+							newServiceEntry.Labels = map[string]string{
+								common.GetWorkloadIdentifier(): fmt.Sprintf("%v", identityId),
+								common.GetEnvKey():             fmt.Sprintf("%v", env),
+							}
+							if seDr.SeDnsPrefix != "" && seDr.SeDnsPrefix != common.Default {
+								newServiceEntry.Labels["dnsPrefix"] = seDr.SeDnsPrefix
+							}
```
aattuluri, line 534: Based on [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names) spec for label names, the above might not work. We can have `dns-prefix`?
Also why labels and not an annotations? Are you using this as filter for querying?

aattuluri (changes requested):
Looks good overall, added comments.
Also some PR checks are failing.

admiral/pkg/clusters/serviceentry.go
vrushalijoshi: Good catch will update the name. I don't see use case where we might need to filter base don dns prefix or GTP, will update these to be annotation. 
Will keep env as label, since identity is today getting added and label, and may be required to filter endpoints, let me know if this is ok?

aattuluri: Yes

vrushalijoshi: cool, updated code based on this

aattuluri (approved):
lgtm

PR #284, opened 2023-02-16
provided an ability to create additional endpoints

shriramsharma:
## Provides an ability to generate additional endpoints for a service

The additional endpoints will be generated based on the `additional_endpoints_suffixes` array passed as the command line param. 
Admiral will use the passed suffixes to generate the endpoints by creating a VirtualService, which will route the corresponding ServiceEntry endpoint

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -304,8 +304,19 @@ func generateProxyVirtualServiceForDependencies(ctx context.Context, remoteRegis
 			if err != nil {
 				return fmt.Errorf("failed generating proxy VirtualService %s due to error: %w", v.Name, err)
 			}
-			log.Infof("successfully generated proxy VirtualService %s", v.Name)
 		}
+	}
+	return nil
+}
+
+func createOrUpdateVirtualService(ctx context.Context, remoteController *RemoteController, virtualService *v1alpha3.VirtualService) error {
+	existingVS, err := remoteController.VirtualServiceController.IstioClient.NetworkingV1alpha3().VirtualServices(common.GetSyncNamespace()).Get(ctx, virtualService.Name, v12.GetOptions{})
+	if err != nil && k8errors.IsNotFound(err) {
+		log.Infof("virtualService %s not found", virtualService.Name)
+	}
+	err = addUpdateVirtualService(ctx, virtualService, existingVS, common.GetSyncNamespace(), remoteController)
+	if err != nil {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


admiral/pkg/clusters/serviceentry.go
```diff
@@ -604,6 +628,122 @@ func AddServiceEntriesWithDr(ctx context.Context, rr *RemoteRegistry, sourceClus
...
+	}
+	log.Infof(LogFormat, "Delete", "VirtualService", defaultVSName, rc.ClusterID, "Success")
+	return nil
+}
+
+// createAdditionalEndpoints creates additional endpoints of service defined in the ServiceEntry.
+// The list suffixes defined in admiralparams.AdditionalEndpointSuffixes will used to generate the endpoints
+func createAdditionalEndpoints(ctx context.Context, serviceEntry *networking.ServiceEntry, namespace string, rc *RemoteController) error {
+
+	additionalEndpointSuffixes := common.GetAdditionalEndpointSuffixes()
```
nirvanagit, line 677: Both create and delete are using common functionality, may be abstract them in a function so both can reuse it.

admiral/pkg/clusters/serviceentry.go
```diff
@@ -309,16 +309,29 @@ func generateProxyVirtualServiceForDependencies(ctx context.Context, remoteRegis
...
+		return nil, fmt.Errorf("no virtualservice found with labels %s", listOptions.LabelSelector)
+	}
+	var result *v1alpha3.VirtualService
+	for _, existingVS := range existingVSList.Items {
+		if isGeneratedByAdmiral(existingVS.Annotations) {
+			result = existingVS
+		}
+	}
+	if result == nil {
+		return nil, fmt.Errorf("no virtualservice found with labels %s and annotation %s=%s", listOptions.LabelSelector, resourceCreatedByAnnotationLabel, resourceCreatedByAnnotationValue)
```
nirvanagit, line 332: Is this a valid scenario where there are no VS created for dependency proxy  in a given cluster?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -309,16 +309,29 @@ func generateProxyVirtualServiceForDependencies(ctx context.Context, remoteRegis
...
-	return nil
+	if existingVSList == nil {
+		return nil, fmt.Errorf("error fetching virtualservice with labels %s", listOptions.LabelSelector)
+	}
+	if len(existingVSList.Items) == 0 {
+		return nil, fmt.Errorf("no virtualservice found with labels %s", listOptions.LabelSelector)
+	}
+	var result *v1alpha3.VirtualService
+	for _, existingVS := range existingVSList.Items {
+		if isGeneratedByAdmiral(existingVS.Annotations) {
```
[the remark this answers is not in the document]
shriramsharma: @nirvanagit , I think the listOptions does not support filtering on annotations

admiral/pkg/clusters/serviceentry.go
shriramsharma: @nirvanagit , this is independent of DependencyProxy. This is the case where the services are part of mesh and communication is not being proxied through a proxy mesh service.

nirvanagit (approved):

PR #285, opened 2023-03-07
generate additional endpoints based on labels set on deployments or rollouts

shriramsharma:


admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -2781,3 +2781,65 @@ func TestGetAdmiralGeneratedVirtualService(t *testing.T) {
...
+		},
+		{
+			name:                           "Given additional endpoint suffixes and labels, when additional endpoint labels filters contains '*', then the func should return true",
+			labels:                         map[string]string{"foo": "bar"},
+			additionalEndpointSuffixes:     []string{"fuzz"},
+			additionalEndpointLabelFilters: []string{"*"},
+			expectedResult:                 true,
+		},
+		{
+			name:                           "Given additional endpoint suffixes and labels, when additional endpoint labels filters contains is not in the rollout/deployment annotation, then the func should return false",
```
nirvanagit, line 2813:
```suggestion
			name:                           "Given additional endpoint suffixes and labels, when label filters do not include any key in labels, then it should return false",
```

admiral/pkg/clusters/serviceentry_test.go
```diff
@@ -2781,3 +2781,65 @@ func TestGetAdmiralGeneratedVirtualService(t *testing.T) {
...
+			labels:                         map[string]string{"foo": "bar"},
+			additionalEndpointSuffixes:     []string{"fuzz"},
+			additionalEndpointLabelFilters: []string{"baz"},
+			expectedResult:                 false,
+		},
+		{
+			name:                           "Given additional endpoint suffixes and labels, when additional endpoint labels filters contains the rollout/deployment annotation, then the func should return true",
```
nirvanagit, line 2820:
```suggestion
			name:                           "Given additional endpoint suffixes and labels, when additional endpoint labels filters contains one of the keys in the labels, then it should return true",
```

nirvanagit (approved):

PR #286, opened 2023-03-10
prevent panic when cluster does not contain rollout OR deployment for given asset

nirvanagit:


admiral/pkg/clusters/util.go
```diff
@@ -14,7 +14,23 @@ import (
...
+	if len(rolloutsByCluster) > 0 && rolloutsByCluster[cluster] != nil {
+		rollout := rolloutsByCluster[cluster]
+		return GetMeshPortsForRollout(cluster, serviceInstance, rollout), rollout.Labels
+	}
+	return nil, nil
+}
+
+func GetMeshPortsForDeployment(clusterName string, destService *k8sV1.Service,
 	destDeployment *k8sAppsV1.Deployment) map[string]uint32 {
 	var meshPorts = destDeployment.Spec.Template.Annotations[common.SidecarEnabledPorts]
```
shriramsharma, line 35: @nirvanagit , should we add a `nil` check here on `destDeployment`?

nirvanagit: Yes absolutely.

shriramsharma (approved):

PR #288, opened 2023-05-04
[MESH-3036] - Set consecutive5xxErrors to 0

vinay-g:
ConsecutiveGatewayErrors in DestinationRule do not take affect the value is higher that consecutive5xxErrors.

In [envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto#envoy-v3-api-field-config-cluster-v3-outlierdetection-consecutive-5xx) consecutive5xxErrors value is set to 5 by default. 

consecutive5xxErrors needs to be set to 0, to explicitly disable the 5XX error outlier detection to kick in.

### DIFF
--- a/admiral/pkg/clusters/handler.go
+++ b/admiral/pkg/clusters/handler.go
@@ -30,6 +30,7 @@ import (
 const (
 	DefaultBaseEjectionTime         int64  = 300
 	DefaultConsecutiveGatewayErrors uint32 = 50
+	DefaultConsecutive5xxErrors     uint32 = 0
 	DefaultInterval                 int64  = 60
 	DefaultHTTP2MaxRequests         int32  = 1000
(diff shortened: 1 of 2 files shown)


itsLucario:
LGTM. 
@vinay-g WDYT about reducing the base_ejection_time and setting max_ejection_time to 300s. So that outlier doesn't become a bottleneck! As of now, we are not setting  max_ejection_time.
Refs:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#ejection-algorithm

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto#envoy-v3-api-field-config-cluster-v3-outlierdetection-max-ejection-time

aattuluri (approved):
lgtm

PR #289, opened 2023-05-04
[MESH-2984] - Adding vm_id to dynamic routing envoyFilter

vinay-g:
vm_id in envoy filter allows the wasm vm to be unique, this will allow multiple filters with different environments to work on the same envoy instance. Example e2e, e2e2 , e2e3. This case before this change would use the same vm in wasm causing issues with config merging.

More info on VM sharing - https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/main/doc/OVERVIEW.md


nirvanagit:
Few questions:
1. What are the different environments in question?
2. When does the merge issue occur?

PR #290, opened 2023-05-08
Add received log line for HandleEventForDeployment

kpharasi:
Add a log line similar to the one in [HandleEventForRollout](https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/clusters/types.go#L629)


adilfulara:
@vinay-g @kpharasi wont this be very chatty? Would it help if we had this at debug and incremented a metric counter instead ?

PR #295, opened 2024-05-17
add interface for getting configurations

nirvanagit:
## Description
Adds an interface for fetching configurations for a given identity, or all the identities in a given cluister.

### DIFF
--- a/admiral/pkg/registry/registry.go
+++ b/admiral/pkg/registry/registry.go
@@ -1,0 +1,10 @@
+package registry
+
+// IdentityConfiguration is an interface to fetch configuration from a registry
+// backend. The backend can provide an API to give configurations per identity,
+// or if given a cluster name, it will provide the configurations for all
+// the identities present in that cluster.
+type IdentityConfiguration interface {
+	GetByIdentity(identityAlias string) error
+	GetByCluster(clusterName string) error
+}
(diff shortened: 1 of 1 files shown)


rtay1188 (approved):

admiral/pkg/registry/registry.go
```diff
@@ -0,0 +1,10 @@
+package registry
+
+// IdentityConfiguration is an interface to fetch configuration from a registry
+// backend. The backend can provide an API to give configurations per identity,
+// or if given a cluster name, it will provide the configurations for all
+// the identities present in that cluster.
+type IdentityConfiguration interface {
+	GetByIdentity(identityAlias string) error
```
vrushalijoshi, line 8: We today use GetByIdentity to get values from cache within admiral codebase, should this be named different may be GetIdentityConfigurationByAlias?

vrushalijoshi (changes requested):

admiral/pkg/registry/registry.go
nirvanagit: Should I rename this to GetByIdentityName?

nirvanagit: Or, given we are already in the IdentityConfiguration interface, it might be okay if we do GetByIdentity, as its a different function.

And if we add GetConfigurationByIdentity, then we would be repeating the interface name.

vrushalijoshi: Sure makes sense, we can use GetByIdentityName, just in case we need another api integration next time for identity

vrushalijoshi (approved):
Looks Good

PR #297, opened 2024-06-13
adding missing util functions

nirvanagit:


rtay1188 (approved):

rtay1188 (changes requested):
Need to add the SWAware feature flag but change it to partitionAware for open source

PR #330, opened 2024-08-24
Mesh 3865 - Adding unique VS name for copy VS feature

vinay-g:
### Checklist

### Description

This change makes sure that the virtual service that gets copied across different clusters does not get over-written. It assumes that the namespace-vsname combination is unique.

Thank you!

### DIFF
--- a/admiral/pkg/clusters/virtualservice_handler.go
+++ b/admiral/pkg/clusters/virtualservice_handler.go
@@ -59,6 +59,7 @@ type SyncVirtualServiceResource func(
 	remoteRegistry *RemoteRegistry,
 	sourceCluster string,
 	syncNamespace string,
+	vsName string,
 ) error
 
 // VirtualServiceHandler responsible for handling Add/Update/Delete events for
@@ -159,6 +160,8 @@ func (vh *VirtualServiceHandler) handleVirtualServiceEvent(ctx context.Context, 
 		return nil
 	}
 
+	vSName := common.GenerateUniqueNameForVS(virtualService.Namespace, virtualService.Name)
+
 	dependentClusters := vh.remoteRegistry.AdmiralCache.CnameDependentClusterCache.Get(spec.Hosts[0]).CopyJustValues()
 	if len(dependentClusters) > 0 {
 		// Add source clusters to the list of clusters to copy the virtual service
@@ -172,6 +175,7 @@ func (vh *VirtualServiceHandler) handleVirtualServiceEvent(ctx context.Context, 
 			vh.remoteRegistry,
 			vh.clusterID,
 			syncNamespace,
+			vSName,
 		)
 		if err != nil {
 			log.Warnf(LogErrFormat, "Sync", common.VirtualServiceResourceType, virtualService.Name, dependentClusters, err.Error()+": sync to dependent clusters will not be retried")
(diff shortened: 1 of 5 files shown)


admiral/pkg/controller/common/common.go
```diff
@@ -702,3 +702,24 @@ func GetMeshPortsHelper(meshPorts string, destService *k8sV1.Service, clusterNam
...
+
+	if syncNamespace == "" && vsName == "" {
+		return ""
+	} else if syncNamespace == "" {
+		return vsName
+	} else if vsName == "" {
+		return syncNamespace
+	}
+
+	newVSName := syncNamespace + "-" + vsName
```
shriramsharma, line 716: is this enough to generate a unique name? Wont we have the same issue if two clients share the same name, then we may overwrite them in the sync namespace?

admiral/pkg/clusters/virtualservice_handler.go
```diff
@@ -397,49 +427,62 @@ func syncVirtualServiceToRemoteCluster(
...
 	}
 	if rc.VirtualServiceController == nil {
-		return fmt.Errorf(LogFormat, "Event", common.VirtualServiceResourceType, virtualService.Name, cluster, "VirtualService controller not initialized for cluster")
+		return fmt.Errorf(LogFormat, "Event", common.VirtualServiceResourceType, vSName, cluster, "VirtualService controller not initialized for cluster")
 	}
+
 	if event == common.Delete {
-		err := rc.VirtualServiceController.IstioClient.NetworkingV1alpha3().VirtualServices(syncNamespace).Delete(ctx, virtualService.Name, metav1.DeleteOptions{})
+		// Best effort delete for existing virtual service with old name
+		_ = rc.VirtualServiceController.IstioClient.NetworkingV1alpha3().VirtualServices(syncNamespace).Delete(ctx, virtualService.Name, metav1.DeleteOptions{})
```
kpharasi, line 450: Why not return an err from here as well?

admiral/pkg/clusters/virtualservice_handler.go
```diff
@@ -397,49 +427,62 @@ func syncVirtualServiceToRemoteCluster(
...
 		exist = nil
 	}
 	if isDeadCluster(err) {
-		ctxLogger.Warnf(LogErrFormat, "Create/Update", common.VirtualServiceResourceType, virtualService.Name, cluster, "dead cluster")
+		ctxLogger.Warnf(LogErrFormat, "Create/Update", common.VirtualServiceResourceType, vSName, cluster, "dead cluster")
 		return nil
 	}
+	err = addUpdateVirtualService(ctxLogger, ctx, virtualService, exist, syncNamespace, rc, remoteRegistry)
+
+	// Best effort delete of existing virtual service with old name
```
kpharasi, line 482: Why are we doing best effort and not returning an error?

admiral/pkg/controller/common/common.go
kpharasi: +1 maybe lets append a random number?

admiral/pkg/clusters/virtualservice_handler.go
vinay-g: Doing best effort as this will eventually always fail once all the vs's are overwritten with new name.

admiral/pkg/clusters/virtualservice_handler.go
vinay-g: We cant gurantee that a VS with the old name will always exist, this code  should be removed after a few months once we over write all the existing vs's with the new ones.

admiral/pkg/controller/common/common.go
```diff
@@ -702,3 +702,24 @@ func GetMeshPortsHelper(meshPorts string, destService *k8sV1.Service, clusterNam
 	}
 	return ports
 }
+
+func GenerateUniqueNameForVS(originNamespace string, vsName string) string {
+
+	if originNamespace == "" && vsName == "" {
```
shriramsharma, line 708: nitpick, you could just make them individual `if` condition rather than `else if`

shriramsharma (approved):

PR #338, opened 2024-09-09
generating ingress DR for VS based routing

shriramsharma:
### Description
Generates the DR which applied to the ingress in order to facilitate VS based cross cluster routing

### DIFF
--- a/admiral/pkg/clusters/virtualservice_routing.go
+++ b/admiral/pkg/clusters/virtualservice_routing.go
@@ -8,6 +8,7 @@ import (
 
 	argo "github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	"github.com/istio-ecosystem/admiral/admiral/pkg/controller/common"
+	"github.com/istio-ecosystem/admiral/admiral/pkg/util"
 	log "github.com/sirupsen/logrus"
 	networkingV1Alpha3 "istio.io/api/networking/v1alpha3"
 	"istio.io/client-go/pkg/apis/networking/v1alpha3"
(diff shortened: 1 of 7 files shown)


admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -16,6 +16,10 @@ import (
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
```
kpharasi, line 19: Let's move this here: https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/util/constants.go?

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -619,3 +620,115 @@ func getMeshHTTPPortForRollout(ports map[string]map[string]uint32) (uint32, erro
...
+			continue
+		}
+
+		for _, drHost := range drHosts {
+			drObj := networkingV1Alpha3.DestinationRule{
+				Host:     drHost,
+				ExportTo: common.GetIngressVSExportToNamespace(),
+				TrafficPolicy: &networkingV1Alpha3.TrafficPolicy{
+					LoadBalancer: &networkingV1Alpha3.LoadBalancerSettings{
+						LbPolicy: &networkingV1Alpha3.LoadBalancerSettings_Simple{
```
vinay-g, line 679: We need to add something below, else we will have issues where AZ distribution will be a problem - 
```
localityLbSetting:
        distribute:
        - from: us-west-2/*
          to:
            us-west-2/*: 100
        - from: us-east-2/*
          to:
            us-west-2/*: 100
```

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -619,3 +620,115 @@ func getMeshHTTPPortForRollout(ports map[string]map[string]uint32) (uint32, erro
...
+	sourceClusterToDRHosts map[string]map[string]string,
+	sourceIdentity string) error {
+
+	if sourceIdentity == "" {
+		return fmt.Errorf("sourceIdentity is empty")
+	}
+
+	san := fmt.Sprintf("%s%s/%s", common.SpiffePrefix, common.GetSANPrefix(), sourceIdentity)
+
+	for sourceCluster, drHosts := range sourceClusterToDRHosts {
```
vinay-g, line 664: Question would it better to create one DR with host like - `"*.sandbox-sandbox-vinayclient2-usw2-qal.svc.cluster.local`

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -619,3 +620,115 @@ func getMeshHTTPPortForRollout(ports map[string]map[string]uint32) (uint32, erro
...
+						},
+					},
+					Tls: &networkingV1Alpha3.ClientTLSSettings{
+						SubjectAltNames: []string{san},
+					},
+				},
+			}
+			drName := fmt.Sprintf("%s-routing-dr", drHost)
+
+			newDR := createDestinationRuleSkeleton(drObj, drName, util.IstioSystemNamespace)
```
vinay-g, line 690: Last time you mentioned VS was it admiral-sync, checking if istio-system will work here

vinay-g (changes requested):

admiral/pkg/clusters/virtualservice_routing.go
shriramsharma: I dont see any reason why we cant do that. ok, let me make this change

admiral/pkg/clusters/virtualservice_routing.go
shriramsharma: @vinay-g , I'll test out but I think last time I tested, the DR either had to be in the service's NS or in the root NS

shriramsharma: @vinay-g , I tested this out and the DR has to apply either in service's own NS or the root NS (istio-system)

vinay-g (approved):

PR #339, opened 2024-09-09
Client (mesh egress) discovery for Jobs and Numaflow (vertices and monovertices)

aattuluri:
### Checklist

### Description
This change introduces client discovery where there is no ingress but only mesh egress required like k8s Jobs, numaflow Vertices and Monovertices.

[Link to related ISSUE]

Thank you!


admiral/pkg/controller/admiral/monovertex.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (d *MonoVertexController) Updated(ctx context.Context, obj interface{}, oldObj interface{}) error {
+	//Not Required, to be handled via off boarding
+	return nil
+}
+
+func addUpdateMonoVertex(j *MonoVertexController, ctx context.Context, obj interface{}) error {
+	monoVertex, ok := obj.(v1alpha1.MonoVertex)
+	if !ok {
+		return nil
```
kpharasi, line 211: return an error if !ok

admiral/pkg/controller/admiral/monovertex.go
```diff
@@ -0,0 +1,268 @@
...
+
+func NewMonoVertexCache() *monoVertexCache {
+	return &monoVertexCache{
+		cache: make(map[string]*MonoVertexEntry),
+		mutex: &sync.Mutex{},
+	}
+}
+
+
+func (p *monoVertexCache) getK8sObjectFromMonoVertex(monoVertex v1alpha1.MonoVertex) *K8sObject{
```
kpharasi, line 49: We probably don't need a pointer receiver here

admiral/pkg/client/loader/kube_loader.go
```diff
@@ -85,6 +86,19 @@ func (loader *KubeClientLoader) LoadKubeClientFromConfig(config *rest.Config) (k
 	return kubernetes.NewForConfig(config)
 }
 
+func (loader *KubeClientLoader) LoadNumaflowClientFromPath(kubeConfigPath string) (numaflow.Interface, error) {
```
kpharasi, line 89: should we have this in a new file called - `numaflow_loader.go`?

admiral/pkg/clusters/handler.go
```diff
@@ -31,6 +32,14 @@ type WeightedService struct {
 	Service *coreV1.Service
 }
 
+type K8sObjectMetadata struct {
```
kpharasi, line 35: Should we move this to: https://github.com/istio-ecosystem/admiral/blob/585c2ea5d9dab5642352f2c8f14794cb1b8006bb/admiral/pkg/controller/common/types.go?

admiral/pkg/controller/admiral/vertex.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (d *VertexController) Updated(ctx context.Context, obj interface{}, oldObj interface{}) error {
+	//Not Required, to be handled via off boarding
+	return nil
+}
+
+func addUpdateVertex(j *VertexController, ctx context.Context, obj interface{}) error {
+	vertex, ok := obj.(v1alpha1.Vertex)
+	if !ok {
+		return nil
```
kpharasi, line 211: return an error if !ok

admiral/pkg/controller/admiral/vertex.go
```diff
@@ -0,0 +1,268 @@
...
+
+func NewVertexCache() *vertexCache {
+	return &vertexCache{
+		cache: make(map[string]*VertexEntry),
+		mutex: &sync.Mutex{},
+	}
+}
+
+
+func (p *vertexCache) getK8sObjectFromVertex(vertex v1alpha1.Vertex) *K8sObject{
```
kpharasi, line 49: We probably don't need a pointer receiver here

admiral/pkg/controller/admiral/job.go
```diff
@@ -0,0 +1,273 @@
...
+	//Not Required, to be handled via off boarding
+	return nil
+}
+
+func addUpdateJob(j *JobController, ctx context.Context, obj interface{}) error {
+	job, ok := obj.(*v12.Job)
+	if !ok {
+		return nil
+	}
+	if !common.ShouldIgnore(job.Annotations, job.Labels) {
```
kpharasi, line 214: Jobs created by cronjobs would also start getting watched by this controller. In the `ShouldIgnore` function we should also ignore jobs which have an `ownerReferences` section. WDYT?

admiral/pkg/client/loader/kube_loader.go
aattuluri: I see istio client loaded in the same file, do you see an advantage in separating it out?

admiral/pkg/clusters/handler.go
aattuluri: Good idea, let me do that.

admiral/pkg/controller/admiral/controller.go
```diff
@@ -80,6 +80,19 @@ type Controller struct {
 	informer  cache.SharedIndexInformer
 }
 
+type K8sObject struct {
```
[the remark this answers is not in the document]
aattuluri: Done

admiral/pkg/clusters/handler.go
aattuluri: Done

admiral/pkg/controller/admiral/job.go
aattuluri: All mesh annotations go under job spec in a cron job, so it should be fine to support this at the job spec/job level.

admiral/pkg/controller/common/common.go
```diff
@@ -703,6 +707,56 @@ func GetMeshPortsHelper(meshPorts string, destService *k8sV1.Service, clusterNam
...
+
+	//if sidecar not injected
+	if annotations[labelSet.DeploymentAnnotation] != "true" { //Not sidecar injected, we don't want to inject
+		return true
+	}
+
+	if annotations[AdmiralIgnoreAnnotation] == "true" {
+		return true
+	}
+
```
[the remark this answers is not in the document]
aattuluri: I deliberately chose to not do namespace as thats not in use currently and we should skip that to save some cycles.

admiral/pkg/controller/admiral/vertex.go
aattuluri: @kpharasi You mean to K8sObject can be returned as value and not reference?

admiral/pkg/controller/common/common.go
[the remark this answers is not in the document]
kpharasi: Got it, makes sense.

admiral/pkg/client/loader/kube_loader.go
kpharasi: Got it, if it has istio client we can keep this here as well. I thought the file is only for k8s client.

admiral/pkg/controller/admiral/vertex.go
kpharasi: I meant we can directly call this function rather than using the `vertexCache` object to call it  as we are not using any field from the `vertexCache` struct.

admiral/pkg/controller/admiral/job.go
kpharasi: Got it, makes sense.

admiral/pkg/controller/admiral/job.go
```diff
@@ -0,0 +1,273 @@
...
+
+func NewJobCache() *jobCache {
+	return &jobCache{
+		cache: make(map[string]*JobEntry),
+		mutex: &sync.Mutex{},
+	}
+}
+
+
+func (p *jobCache) getK8sObjectFromJob(job *v12.Job) *K8sObject{
```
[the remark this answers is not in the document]
aattuluri: Done

admiral/pkg/clusters/handler.go
```diff
@@ -323,3 +325,94 @@ func GetServiceWithSuffixMatch(suffix string, services []*coreV1.Service) string
 	}
 	return ""
 }
+
+func HandleEventForClientDiscovery(ctx context.Context, event admiral.EventType, obj *common.K8sObject,
```
shriramsharma, line 329: can this func be moved to `clientdiscovery_handler.go`?

admiral/pkg/clusters/handler.go
```diff
@@ -323,3 +325,94 @@ func GetServiceWithSuffixMatch(suffix string, services []*coreV1.Service) string
...
+		return nil
+	}
+
+	ctx = context.WithValue(ctx, "clusterName", clusterName)
+	ctx = context.WithValue(ctx, "eventResourceType", obj.Type)
+
+	if remoteRegistry.AdmiralCache != nil {
+
+		UpdateIdentityClusterCache(remoteRegistry, globalIdentifier, clusterName)
+
```
shriramsharma, line 362: I think we should return an error if `remoteRegistry.AdmiralCache == nil` and not move ahead because if `UpdateIdentityClusterCache` is not called then calling Added of dependency controller might not do anything.

aattuluri: Good catch!

admiral/pkg/clusters/handler.go
aattuluri: Done

admiral/pkg/clusters/handler.go
```diff
@@ -323,3 +325,94 @@ func GetServiceWithSuffixMatch(suffix string, services []*coreV1.Service) string
...
+		return nil
+	}
+	ctxLogger := common.GetCtxLogger(ctx, globalIdentifier, "")
+	if commonUtil.IsAdmiralReadOnly() {
+		ctxLogger.Infof(common.CtxLogFormat, event, "", "", "", "processing skipped as Admiral is in Read-only mode")
+		return nil
+	}
+
+	// Should not return early here for TrafficConfig persona, as cache should build up during warm up time
+	if IsCacheWarmupTime(remoteRegistry) && !common.IsPersonaTrafficConfig() {
```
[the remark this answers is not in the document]
aattuluri: Done

admiral/pkg/clusters/dependency_handler.go
```diff
@@ -87,15 +87,17 @@ func isIdentityMeshEnabled(identity string, remoteRegistry *RemoteRegistry) bool
...
 	updatedDependency *v1.Dependency, remoteRegistry *RemoteRegistry) ([]string, bool) {
 	updatedDestinations := make([]string, 0)
 	existingDestination := remoteRegistry.AdmiralCache.SourceToDestinations.Get(updatedDependency.Spec.Source)
-
 	var nonMeshEnabledExists bool
 	lookup := make(map[string]bool)
-	for _, dest := range existingDestination {
-		lookup[dest] = true
+	//if this is an update, build a look up table to process only the diff
+	if eventType == admiral.Update {
```
kpharasi, line 97: Why was this added?

admiral/pkg/controller/admiral/job.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return &jobController, nil
+}
+
+func (d *JobController) Added(ctx context.Context, obj interface{}) error {
+	return addUpdateJob(d, ctx, obj)
+}
+
+func (d *JobController) Updated(ctx context.Context, obj interface{}, oldObj interface{}) error {
+	//Not Required, to be handled via off boarding
```
kpharasi, line 192: We just want to skip Update? I think this comment is valid for only delete. Same for the other 2.

aattuluri: An Update to the job is kind of not relevant as we just need the Job to be registered as a client when its created.

admiral/pkg/clusters/dependency_handler.go
aattuluri: `Add` event for dependency record was never processed. Meaning all calls to this method would assume dependency record update (get from cache, compare that with updated object). Adding this will make sure that Add event will consider all destinations to be processed.

kpharasi: Cool, thanks!

kpharasi (approved):

PR #340, opened 2024-09-10
fix casting of slices of services

rtay1188:
### Checklist

- [x] I've read and agree to the project's contribution guidelines.
- [x] I'm requesting to **pull a topic/feature/bugfix branch**.
- [x] I checked that my code additions will pass code linting checks and unit tests.
- [x] I updated unit and integration tests (if applicable).
- [x] I'm ready to notify the team of this contribution.

### Description
Previously, we were attempting to cast []*v1.Service to metav1.Object. This is because the cached services are stored as slices, as each identity could have more than one service. Instead, we check the service that we got the event for against each of the services in the slice to pick the right one in the cache to cast to metav1.Object.

Thank you!

### DIFF
--- a/admiral/pkg/controller/admiral/controller.go
+++ b/admiral/pkg/controller/admiral/controller.go
@@ -455,6 +455,10 @@ func checkIfResourceVersionHasIncreased(ctxLogger *logrus.Entry, ctx context.Con
 }
 
 func shouldRetry(ctxLogger *logrus.Entry, ctx context.Context, obj interface{}, delegator Delegator) bool {
+	var (
+		latestObjMeta metav1.Object
+		latestOk      bool
+	)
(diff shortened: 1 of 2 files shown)


admiral/pkg/controller/admiral/controller.go
```diff
@@ -465,8 +466,11 @@ func shouldRetry(ctxLogger *logrus.Entry, ctx context.Context, obj interface{},
 			return true
 		}
 		latestObjMeta, latestOk := objFromCache.(metav1.Object)
+		if servicesSlice, servicesSliceOk := objFromCache.([]*v1.Service); servicesSliceOk {
```
kpharasi, line 469: Let's use a switch statement for this

kpharasi (approved):

PR #346, opened 2024-09-17
added a cluster allowlist param for vs based routing

shriramsharma:
### Description
Added a cluster allow list to specify on which source clusters VS based routing should be enabled.

### DIFF
--- a/admiral/cmd/admiral/cmd/root.go
+++ b/admiral/cmd/admiral/cmd/root.go
@@ -254,6 +254,7 @@ func GetRootCmd(args []string) *cobra.Command {
 
 	// Flags pertaining to VS based routing
 	rootCmd.PersistentFlags().BoolVar(&params.EnableVSRouting, "enable_vs_routing", false, "Enable/Disable VS Based Routing")
+	rootCmd.PersistentFlags().StringArrayVar(&params.VSRoutingEnabledClusters, "vs_routing_enabled_clusters", []string{}, "The source clusters to enable VS based routing on")
 	rootCmd.PersistentFlags().StringArrayVar(&params.VSRoutingGateways, "vs_routing_gateways", []string{}, "The PASSTHROUGH gateways to use for VS based routing")
 	rootCmd.PersistentFlags().StringArrayVar(&params.IngressVSExportToNamespaces, "ingress_vs_export_to_namespaces", []string{"istio-system"}, "List of namespaces where the ingress VS should be exported")
 	rootCmd.PersistentFlags().StringVar(&params.IngressLBPolicy, "ingress_lb_policy", "round_robin", "loadbalancer policy for ingress destination rule (round_robin/random/passthrough/least_request)")
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -759,7 +759,7 @@ func modifyServiceEntryForNewServiceOrPod(
 			remoteRegistry.AdmiralCache.DependencyNamespaceCache.Put(val, serviceInstance[appType[sourceCluster]].Namespace, localFqdn, cnames)
 		}
 
-		if common.IsVSBasedRoutingEnabled() {
+		if common.DoVSRoutingForCluster(sourceCluster) {
 			// Discovery phase: This is where we build a map of all the svc.cluster.local destinations
 			// for the identity's source cluster. This map will contain the RouteDestination of all svc.cluster.local
 			// endpoints.
@@ -779,7 +779,6 @@ func modifyServiceEntryForNewServiceOrPod(
 					"No RouteDestinations generated for VS based routing ")
 			} else {
 				sourceClusterToDestinations[sourceCluster] = destinations
-				// Get the hosts to populate the DR
 				drHost := fmt.Sprintf("*.%s.%s", deploymentOrRolloutNS, common.DotLocalDomainSuffix)
 				sourceClusterToDRHosts[sourceCluster] = map[string]string{
 					deploymentOrRolloutNS + common.DotLocalDomainSuffix: drHost,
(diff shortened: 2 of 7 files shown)


Punakshi (approved):

admiral/pkg/controller/common/config.go
```diff
@@ -438,10 +438,18 @@ func GetIngressVSExportToNamespace() []string {
...
 
-func IsVSBasedRoutingEnabled() bool {
+func DoVSRoutingForCluster(cluster string) bool {
 	wrapper.RLock()
 	defer wrapper.RUnlock()
-	return wrapper.params.EnableVSRouting
+	if !wrapper.params.EnableVSRouting {
+		return false
+	}
+	for _, c := range wrapper.params.VSRoutingEnabledClusters {
```
adilfulara, line 447: If `EnableVSRouting` is `true` and `VSRoutingEnabledClusters` is "[]", does that mean no clusters should get the VS? 

What is the correct way to enable it across the board without adding clusters to the list?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -759,7 +759,7 @@ func modifyServiceEntryForNewServiceOrPod(
 			remoteRegistry.AdmiralCache.DependencyNamespaceCache.Put(val, serviceInstance[appType[sourceCluster]].Namespace, localFqdn, cnames)
 		}
 
-		if common.IsVSBasedRoutingEnabled() {
+		if common.DoVSRoutingForCluster(sourceCluster) {
```
adilfulara, line 762: Is this discover phase, could we rename to `DiscoverVS(sourceCluster)`?

admiral/pkg/clusters/serviceentry.go
```diff
@@ -802,28 +801,26 @@ func modifyServiceEntryForNewServiceOrPod(
 		ctxLogger.Infof(common.CtxLogFormat, "updateRegistryConfigForClusterPerEnvironment", deploymentOrRolloutName, deploymentOrRolloutNS, "", "done writing")
 		return nil, nil
 	}
-	// If VS based routing is enabled, then generate VirtualServices for the source cluster's ingress
-	// This is done after the ServiceEntries are created for the source cluster
-	if common.IsVSBasedRoutingEnabled() {
-		// Writing phase: We update the base ingress virtualservices with the RouteDestinations
-		// gathered during the discovery phase and write them to the source cluster
-		err := addUpdateVirtualServicesForSourceIngress(ctx, ctxLogger, remoteRegistry, sourceClusterToDestinations)
```
adilfulara, line 810: Should we rename to `WriteVSForIngress`?
Is `SourceIngress` or `DestinationIngress` of any concern?

Also why do we have global functions? Shouldn't a source cluster have this API to write to a destination?

adilfulara (commented):
overall looks good. added comments. lmk wdyt?

admiral/pkg/controller/common/config.go
vinay-g: +1

admiral/cmd/admiral/cmd/root.go
```diff
@@ -254,6 +254,7 @@ func GetRootCmd(args []string) *cobra.Command {
 
 	// Flags pertaining to VS based routing
 	rootCmd.PersistentFlags().BoolVar(&params.EnableVSRouting, "enable_vs_routing", false, "Enable/Disable VS Based Routing")
```
vinay-g, line 256: This probably needs to be a string with enabled/disabled/all or something

vinay-g (changes requested):

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -454,15 +454,16 @@ func addUpdateVirtualServicesForSourceIngress(
...
+	for sourceCluster, destination := range sourceClusterToDestinations {
 
-	if len(sourceClusterToDestinations) == 0 {
-		return fmt.Errorf("no route destination found for the ingress virtualservice")
-	}
+		if !common.DoVSRoutingForCluster(sourceCluster) {
+			continue
+		}
+
+		if remoteRegistry == nil {
```
vinay-g, line 463: Can this check be done only once ?

admiral/pkg/controller/common/config.go
shriramsharma: @adilfulara , I'll be adding a check for `*`. If that exists in the cluster list, then it will be enabled for all clusters.

admiral/pkg/clusters/serviceentry.go
shriramsharma: @adilfulara , this just a feature flag for enabling/disabling VS based routing

admiral/pkg/clusters/serviceentry.go
shriramsharma: changing the name to `addUpdateVirtualServicesForIngress`
This is not exactly a global func, this func is accessible from within the package.

admiral/cmd/admiral/cmd/root.go
shriramsharma: @vinay-g would this work? https://github.com/istio-ecosystem/admiral/pull/346#discussion_r1764429203

admiral/pkg/clusters/serviceentry.go
adilfulara: true. to clarify i meant it should be been a method on a specific struct on the source cluster. but it seems we dont have the struct at this time.

admiral/pkg/controller/common/config_test.go
```diff
@@ -56,6 +56,76 @@ func setupForConfigTests() {
...
+				"Then it should return false",
+			cluster:                   "cluster1",
+			enableVSRouting:           true,
+			enableVSRoutingForCluster: []string{"cluster1"},
+			expected:                  true,
+		},
+		{
+			name: "Given enableVSRouting is true, and all VS routing is enabled in all clusters using '*'" +
+				"When DoVSRoutingForCluster is called" +
+				"Then it should return false",
```
adilfulara, line 108: `Then it should return true` :)

adilfulara (approved):

vinay-g (approved):

PR #362, opened 2024-12-09
MESH-5878: Admiral does not create envoy filter on client onboardings

adilfulara:
Task
- Admiral should process new client onboardings for a service with a RoutingPolicy

### Description
- Introduce RoutingPolicyProcessor interface to contain the process() api in order to expose it

Thank you!

### DIFF
--- a/admiral/pkg/clusters/dependency_handler.go
+++ b/admiral/pkg/clusters/dependency_handler.go
@@ -106,7 +106,9 @@ func (d *ProcessDestinationService) Process(ctx context.Context, dependency *v1.
 	}
 
 	destinations, hasNonMeshDestination := getDestinationsToBeProcessed(eventType, dependency, remoteRegistry)
-	log.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("found %d new destinations: %v", len(destinations), destinations))
+	id := common.GetIdentity(dependency.Labels, nil)
+	ctxLogger := common.GetCtxLogger(ctx, id, "-")
+	ctxLogger.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("found %d new destinations: %v", len(destinations), destinations))
 
 	// find source cluster for source identity
 	sourceClusters := remoteRegistry.AdmiralCache.IdentityClusterCache.Get(dependency.Spec.Source)
@@ -115,7 +117,7 @@ func (d *ProcessDestinationService) Process(ctx context.Context, dependency *v1.
 		// the rollout/deployment event hasn't gone through yet.
 		// This can be ignored, and not be added back to the dependency controller queue
 		// because it will be processed by the rollout/deployment controller
-		log.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("identity: %s, does not have any clusters. Skipping calling modifySE", dependency.Spec.Source))
+		ctxLogger.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("identity: %s, does not have any clusters. Skipping calling modifySE", dependency.Spec.Source))
 		return nil
 	}
 
--- a/admiral/pkg/clusters/dependency_handler.go
+++ b/admiral/pkg/clusters/dependency_handler.go
@@ -74,13 +74,13 @@ func (dh *DependencyHandler) HandleDependencyRecord(ctx context.Context, obj *v1
 		// This will be re-queued and retried
 		return handleDepRecordErrors
 	}
-	remoteRegistry.AdmiralCache.SourceToDestinations.put(obj)
-
 	// process routing policies.
 	err = dh.RoutingPolicyProcessor.ProcessDependency(ctx, eventType, obj)
 	if err != nil {
 		log.Errorf(LogErrFormat, string(eventType), common.DependencyResourceType, obj.Name, "", fmt.Sprintf("error processing routing policies %v", err))
 	}
+
+	remoteRegistry.AdmiralCache.SourceToDestinations.put(obj)
 
 	return handleDepRecordErrors
 }
--- a/admiral/pkg/clusters/dependency_handler.go
+++ b/admiral/pkg/clusters/dependency_handler.go
@@ -61,12 +61,6 @@ func (dh *DependencyHandler) HandleDependencyRecord(ctx context.Context, obj *v1
 
 	var handleDepRecordErrors error
 
-	// process routing policies.
-	err = dh.RoutingPolicyProcessor.ProcessDependency(ctx, eventType, obj)
-	if err != nil {
-		log.Errorf(LogErrFormat, string(eventType), common.DependencyResourceType, obj.Name, "", fmt.Sprintf("error processing routing policies %v", err))
-	}
-
 	// Generate SE/DR/VS for all newly added destination services in the source's cluster
 	err = dh.DestinationServiceProcessor.Process(ctx,
 		obj,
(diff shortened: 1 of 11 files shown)


admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,201 @@ import (
...
+					err = common.AppendError(err, err1)
+				} else {
+					log.Infof(LogFormat, eventType, "routingpolicy	", newRP.Name, remoteController.ClusterID, "created envoyfilters")
+				}
+			}
+			id := common.GetRoutingPolicyIdentity(newRP)
+			if oldRP != nil {
+				oldEnv := common.GetRoutingPolicyEnv(oldRP)
+				r.RemoteRegistry.AdmiralCache.RoutingPolicyCache.Delete(id, oldEnv, oldRP.Name)
+			}
```
shriramsharma, line 60: Why are we deleting from the cache?

admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,201 @@ import (
...
+			env := common.GetRoutingPolicyEnv(newRP)
+			r.RemoteRegistry.AdmiralCache.RoutingPolicyCache.Put(id, env, newRP.Name, newRP)
+		}
+	}
+	return err
+}
+
+func (r *RoutingPolicyService) ProcessDependency(ctx context.Context, eventType admiral.EventType, dependency *v1.Dependency) error {
+	//ns := getEnvoyFilterNamespace()
```
shriramsharma, line 69: can we remove this?

admiral/pkg/clusters/routingpolicy_handler.go
adilfulara: We are deleting the old resource as part of the update event. Should we be retaining older resources?

admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,201 @@ import (
...
+	newDestinations, _ := getDestinationsToBeProcessed(eventType, dependency, r.RemoteRegistry)
+
+	// for each destination, identify the newRP.
+	for _, dependent := range newDestinations {
+		// if the dependent is in the mesh, then get the newRP
+		policies := r.RemoteRegistry.AdmiralCache.RoutingPolicyCache.GetForIdentity(dependent)
```
shriramsharma, line 75: sorry please correct me if I'm wrong, but why are we trying to get routingPolicies of the destination/client?

admiral/pkg/clusters/routingpolicy_handler.go
adilfulara: done.

admiral/pkg/clusters/routingpolicy_handler.go
adilfulara: `newDestinations` is the new service that the client onboarded to. So we need to get all routing policies associated with the destination so we can apply it at the client side.

admiral/pkg/clusters/routingpolicy_handler.go
vinay-g: Since we dont have routing policies in ppd or production we may not need this. Plus cache ( in memory) gets populated every time admiral restarts.

admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,200 @@ import (
...
-	identityCache map[string]*v1.RoutingPolicy
-	mutex         *sync.Mutex
+func (r *RoutingPolicyService) ProcessAddOrUpdate(ctx context.Context, eventType admiral.EventType, newRP *v1.RoutingPolicy, oldRP *v1.RoutingPolicy, dependents map[string]string) error {
+	var err error
+	for _, remoteController := range r.RemoteRegistry.remoteControllers {
+		if !common.DoRoutingPolicyForCluster(remoteController.ClusterID) {
+			log.Warnf(LogFormat, eventType, "routingpolicy", newRP.Name, remoteController.ClusterID, "RoutingPolicy disabled for cluster")
+			continue
+		}
+		for _, dependent := range dependents {
```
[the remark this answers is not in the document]
adilfulara: We could definitely benefit from usage of go-routines but i chose not do this for following reasons
- currently there are no such use cases.
- each client to service mapping will come in as a separate update due to the DependencyRecord CRD, there wont be much benefit from usage of goroutine usage.

LMK what you prefer.

admiral/pkg/clusters/routingpolicy_handler.go
adilfulara: the reason i am deleting is if the `env` label is changed the cache will be corrupted which could cause problems for subsequent updates. ( ex: change is admiral.io/env label value )

admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,200 @@ import (
...
+			err := r.ProcessAddOrUpdate(ctx, eventType, rp, nil, map[string]string{dependent: dependency.Spec.Source})
+			if err != nil {
+				log.Errorf(LogErrFormat, eventType, "routingpolicy", rp.Name, "",
+					fmt.Sprintf("failed to process routing policy for new destination=%s in delta update", dependent))
+				return err
+			}
+			log.Infof(LogFormat, eventType, "routingpolicy", rp.Name, "",
+				fmt.Sprintf("finished processing routing policy for new destination=%s in delta update", dependent))
+		}
+		//}
```
shriramsharma, line 85: nitpick, can this be removed?

admiral/pkg/clusters/routingpolicy_handler.go
shriramsharma: @adilfulara is it ok to refresh the map if there are error in the above loop? For example, what if all the calls to `createOrUpdateEnvoyFilter` return failure. Is it ok to mutate the map?

admiral/pkg/clusters/routingpolicy_handler.go
```diff
@@ -15,51 +16,200 @@ import (
...
+		if rc != nil {
+			if filterMap, ok := clusterIdFilterMap[rc.ClusterID]; ok {
+				for filter, filterNs := range filterMap {
+					log.Infof(LogFormat, eventType, "envoyfilter", filter, rc.ClusterID, "deleting")
+					err1 := rc.RoutingPolicyController.IstioClient.NetworkingV1alpha3().EnvoyFilters(filterNs).Delete(ctx, filter, metaV1.DeleteOptions{})
+					if err1 != nil {
+						log.Errorf(LogErrFormat, eventType, "envoyfilter", filter, rc.ClusterID, err1)
+						err = common.AppendError(err, err1)
+					} else {
+						log.Infof(LogFormat, eventType, "envoyfilter", filter, rc.ClusterID, "deleting from cache")
```
shriramsharma, line 115: this log line message might be inaccurate?

admiral/pkg/clusters/routingpolicy_handler.go
adilfulara: done. good catch. fixed it.

admiral/pkg/clusters/dependency_handler.go
```diff
@@ -99,7 +106,9 @@ func (d *ProcessDestinationService) Process(ctx context.Context, dependency *v1.
 	}
 
 	destinations, hasNonMeshDestination := getDestinationsToBeProcessed(eventType, dependency, remoteRegistry)
-	log.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("found %d new destinations: %v", len(destinations), destinations))
+	id := common.GetIdentity(dependency.Labels, nil)
+	ctxLogger := common.GetCtxLogger(ctx, id, "-")
+	ctxLogger.Infof(LogFormat, string(eventType), common.DependencyResourceType, dependency.Name, "", fmt.Sprintf("found %d new destinations: %v", len(destinations), destinations))
```
shriramsharma, line 111: are these needed now? I guess you added them for debugging right?

PR #366, opened 2024-12-13
fix duplicate vs issue in vs based routing

shriramsharma:
fix duplicate vs issue in vs based routing

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -824,7 +824,7 @@ func modifyServiceEntryForNewServiceOrPod(
 	// VS Based Routing
 	// Writing phase: We update the base ingress virtualservices with the RouteDestinations
 	// gathered during the discovery phase and write them to the source cluster
-	err = addUpdateVirtualServicesForIngress(ctx, ctxLogger, remoteRegistry, sourceClusterToDestinations)
+	err = addUpdateVirtualServicesForIngress(ctx, ctxLogger, remoteRegistry, sourceClusterToDestinations, cname)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


admiral/pkg/clusters/serviceentry.go
```diff
@@ -824,7 +824,7 @@ func modifyServiceEntryForNewServiceOrPod(
 	// VS Based Routing
 	// Writing phase: We update the base ingress virtualservices with the RouteDestinations
 	// gathered during the discovery phase and write them to the source cluster
-	err = addUpdateVirtualServicesForIngress(ctx, ctxLogger, remoteRegistry, sourceClusterToDestinations)
+	err = addUpdateVirtualServicesForIngress(ctx, ctxLogger, remoteRegistry, sourceClusterToDestinations, cname)
```
adilfulara, line 827: since a vs can have multiple hosts, would it be advisable to use an array of `cnames`?

admiral/pkg/clusters/virtualservice_routing_test.go
```diff
@@ -171,6 +171,7 @@ func TestAddUpdateVirtualServicesForIngress(t *testing.T) {
 	testCases := []struct {
 		name                        string
```
adilfulara, line 172: nitpick

i prefer to have a simple test case struct 
```go
struct {
  name string
  args args
  want want
}
```

This way it is pretty clear what the inputs and outputs are.
Not super important though.

admiral/pkg/clusters/serviceentry.go
shriramsharma: @adilfulara this is actually the vs name and I'm just reusing the cname thats being used in other places where cname variable name makes sense. I can create another variable or add a comments describing the intention of this variable in this context.

vinay-g (approved):

PR #375, opened 2025-01-07
MESH-6000: ignore argo tracking tags during VS copy

rtay1188:
### Checklist

- [x] I've read and agree to the project's contribution guidelines.
- [x] I'm requesting to **pull a topic/feature/bugfix branch**.
- [x] I checked that my code additions will pass code linting checks and unit tests.
- [x] I updated unit and integration tests (if applicable).
- [x] I'm ready to notify the team of this contribution.

### Description
Adds an ignore list for tags to be ignored during VS copy or other operations. Default is filled with argo tracking tags.


admiral/pkg/clusters/virtualservice_handler.go
```diff
@@ -529,6 +529,12 @@ func addUpdateVirtualService(
 		skipAddingExportTo = true
 	}
 
+	// remove ignored labels and annotations from NewCopy (deleting on nil map or nonexistent keys is a no-op)
+	for _, ignored := range common.GetIgnoreLabelsAnnotations() {
+		delete(newCopy.Labels, ignored)
+		delete(newCopy.Annotations, ignored)
+	}
```
vinay-g, line 536: Can you double check if we need the logic here - 
 https://github.com/istio-ecosystem/admiral/blob/5b6c7b99874fadd1ac65808d52fc3df265cadeab/admiral/pkg/clusters/virtualservice_handler.go#L578-L579

rtay1188: Doesn't need the logic there since newCopy already had the labels/annotations modified before the if block at line 533. Similarly the retry function takes the newCopy which was modified so it doesn't need the logic in the retry function either.

admiral/cmd/admiral/cmd/root.go
```diff
@@ -240,11 +240,11 @@ func GetRootCmd(args []string) *cobra.Command {
...
 	rootCmd.PersistentFlags().Int64Var(&params.DefaultWarmupDurationSecs, "default_warmup_duration_in_seconds", 45, "The default value for the warmupDurationSecs to be used on Destination Rules created by admiral")
-
 	rootCmd.PersistentFlags().BoolVar(&params.EnableGenerationCheck, "enable_generation_check", true, "Enable/Disable Generation Check")
 	rootCmd.PersistentFlags().BoolVar(&params.EnableIsOnlyReplicaCountChangedCheck, "enable_replica_count_check", false, "Enable/Disable Replica Count Check")
 	rootCmd.PersistentFlags().BoolVar(&params.ClientInitiatedProcessingEnabled, "client_initiated_processing_enabled", true, "Enable/Disable Client Initiated Processing")
 	rootCmd.PersistentFlags().BoolVar(&params.PreventSplitBrain, "prevent_split_brain", true, "Enable/Disable Explicit Split Brain prevention logic")
+	rootCmd.PersistentFlags().StringSliceVar(&params.IgnoreLabelsAnnotationsList, "ignore_labels_annotations_list", []string{"applications.argoproj.io/app-name", "app.kubernetes.io/instance", "argocd.argoproj.io/tracking-id"}, "Labels and annotations that should not be preserved during VS copy")
```
adilfulara, line 247: If this Argo specific then should we indicate that in the name?
If not, then do we need to apply this treatment to all CRD's?

rtay1188: I intended it to be generic in case we need to do anything similar in the future for non-argo use case, but currently yes it is Argo specific. It only needs to be applied to VS, not all CRDs, but I didn't specify that in the name so that it can be more generic for future use case.

adilfulara: Ok it is a nitpick.
If the treatment is not CRD agnostic, then we should not be using generic names. As a new developer, this flag does not tell me it applies to only specific CRDs.

@vinay-g @shriramsharma i think we musth add a section in the README on all the flags available in Admiral.

rtay1188: I've switched the flag to be VS specific

PR #382, opened 2025-02-03
vs routing in-cluster rollout flags

shriramsharma:
vs routing in-cluster rollout flags

### DIFF
--- a/admiral/pkg/controller/common/config_test.go
+++ b/admiral/pkg/controller/common/config_test.go
@@ -308,6 +308,146 @@ func TestIsSlowStartEnabledForCluster(t *testing.T) {
 
 }
 
+func TestDoVSRoutingInClusterForCluster(t *testing.T) {
+	p := AdmiralParams{}
+
+	testCases := []struct {
+		name                                string
+		cluster                             string
+		enableVSRoutingInCluster            bool
+		enabledVSRoutingInClusterForCluster []string
+		expected                            bool
+	}{
+		{
+			name: "Given enableVSRoutingInCluster is false, enabledVSRoutingInClusterForCluster is empty" +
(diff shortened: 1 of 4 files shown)


virajrk (approved):

admiral/pkg/controller/common/config_test.go
```diff
@@ -308,6 +308,146 @@ func TestIsSlowStartEnabledForCluster(t *testing.T) {
...
+			identity:                               "testIdentity1",
+			enableVSRoutingInCluster:               true,
+			enabledVSRoutingInClusterForIdentities: []string{"*"},
+			expected:                               true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			p.EnableVSRoutingInCluster = tc.enableVSRoutingInCluster
```
adilfulara, line 440: https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/clusters/handler_test.go#L26
https://github.com/istio-ecosystem/admiral/blob/master/admiral/pkg/clusters/clientdiscovery_handler_test.go#L249

should we use this approach for initializing admiral  instead of manually modifying struct?

adilfulara: it is a nit pick so feel free to omit.

adilfulara (approved):

vrushalijoshi (approved):
looks good, just the commits need to be signed off before merge.

admiral/pkg/controller/common/config_test.go
shriramsharma: @adilfulara , thats a good point but initAdmiral func does a lot more than what I need. Just wanted to keep it simple and set params that I need.

PR #386, opened 2025-02-11
clean up old code and update configwrite

rtay1188:
Cleaning up old state syncer code now that new code will be in the handlers rather than modifySE.
Also updating the configwriter - these will be removed later but leaving it in for now so that operator work can reference it.


admiral/cmd/admiral/cmd/root.go
```diff
@@ -244,7 +244,7 @@ func GetRootCmd(args []string) *cobra.Command {
 	rootCmd.PersistentFlags().BoolVar(&params.EnableSWAwareNSCaches, "enable_sw_aware_ns_caches", false, "Enable/Disable SW Aware NS Caches")
 	rootCmd.PersistentFlags().Int64Var(&params.DefaultWarmupDurationSecs, "default_warmup_duration_in_seconds", 45, "The default value for the warmupDurationSecs to be used on Destination Rules created by admiral")
 	rootCmd.PersistentFlags().BoolVar(&params.EnableGenerationCheck, "enable_generation_check", true, "Enable/Disable Generation Check")
-	rootCmd.PersistentFlags().BoolVar(&params.EnableIsOnlyReplicaCountChangedCheck, "enable_replica_count_check", true, "Enable/Disable Replica Count Check")
```
rtay1188, line 247: In previous PR I accidentally changed this to true, changing it back to false here.

Punakshi (approved):

PR #393, opened 2025-03-05
moved exportTo to generateInClusterVS func

shriramsharma:
moved exportTo to generateInClusterVS func


admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -524,6 +527,13 @@ func generateVirtualServiceForIncluster(
 
 	virtualService.Name = vsName + "-incluster-vs"
 
+	// Add the exportTo namespaces to the virtual service
+	if common.EnableExportTo(vsName) {
```
vrushalijoshi, line 531: does this need to be behind a feature flag?

vrushalijoshi: ok got it

vrushalijoshi (approved):

PR #395, opened 2025-03-05
Update in-cluster VS failover configuration

vrushalijoshi:
Updating the in-cluster VS with failover configuration

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -172,6 +172,8 @@ func modifyServiceEntryForNewServiceOrPod(
 
 		// Holds the VS destinations for the TLSRoutes
 		sourceClusterToDestinations = make(map[string]map[string][]*vsrouting.RouteDestination)
+		// Holds the VS destinations for the HTTPRoutes
+		sourceClusterToInClusterDestinations = make(map[string]map[string][]*vsrouting.RouteDestination)
 		// Holds the DR hosts (*.svc.cluster.local) used for VS based routing
 		sourceClusterToDRHosts = make(map[string]map[string]string)
 		// Holds the source cluster to namespace mapping from where the event is received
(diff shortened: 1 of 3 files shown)


admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -883,21 +905,61 @@ func getDestinationsForGTPDNSPrefixes(
...
+								Number: 80,
+							}
+						} else if weights[sourceClusterLocality] != 100 {
+							if rd.Weight != 0 {
+								weight := (float32(rd.Weight) / 100) * float32(weights[sourceClusterLocality])
+								rd.Weight = int32(weight)
+							} else {
+								rd.Weight = weights[sourceClusterLocality]
+							}
+							remoteRD = getRouteDestination(routeHost, 80, 100-weights[sourceClusterLocality])
```
shriramsharma, line 952: hey @vrushalijoshi , sorry if I misunderstood this. In this case, if the `sourceClusterLocality=west` and its weight is set to 20%, that means that 20% of the traffic should go to `.local`. But it seems it is doing the opposite. Correct me if I'm wrong though.

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -883,21 +905,61 @@ func getDestinationsForGTPDNSPrefixes(
...
 			}
+
 			newRD, err := copyRouteDestinations(routeDestinations)
 			if err != nil {
 				return nil, err
 			}
+
+			if len(weights) > 0 && updateWeights {
+				for _, rd := range newRD {
+					if strings.HasSuffix(rd.Destination.Host, "svc.cluster.local") {
```
shriramsharma, line 939: nitpick,  can you please use `common.DotLocalDomainSuffix`?

admiral/pkg/clusters/virtualservice_routing.go
vrushalijoshi: Hey, it will update the 20% traffic to route destination (rd) which is basically updating the ".local" endpoint( in this case west). For east a new endpoint with ".global" would be created with 80%

shriramsharma: oh sorry it makes sense. I missed that

admiral/pkg/clusters/virtualservice_routing.go
vrushalijoshi: updated it

admiral/pkg/clusters/serviceentry.go
```diff
@@ -770,44 +772,101 @@ func modifyServiceEntryForNewServiceOrPod(
...
+			}
+			sourceClusterToDestinations[sourceCluster] = ingressDestinations
+		} else {
+			ctxLogger.Infof(common.CtxLogFormat, "processGTPAndAddWeightsByCluster",
+				"", "", sourceCluster,
+				"Discovery phase: Skipped GTP processing as VS based routing disabled for cluster")
+		}
+
+		//Discovery Phase: process in-cluster routing destinations with dns prefixes and weights based on GTP
+		if common.DoVSRoutingInClusterForCluster(sourceCluster) {
```
shriramsharma, line 851: @vrushalijoshi , can you please add this check as well here? `common.DoVSRoutingInClusterForIdentity(sourceIdentity)`

shriramsharma (approved):

PR #397, opened 2025-03-10
added rollback for incluster vs

shriramsharma:
added rollback for incluster vs


admiral/pkg/clusters/serviceentry.go
```diff
@@ -897,6 +897,18 @@ func modifyServiceEntryForNewServiceOrPod(
 		}
 	}
 
+	// VS Based Routing - In Cluster
+	// Rollback if this identity exists in vs_routing_in_cluster_disabled_identities list
+	// Rollback for all identities if the current identity's source clusters are in
+	// vs_routing_in_cluster_disabled_clusters list
+	err = performInVSRoutingRollback(
```
vrushalijoshi, line 904: do not execute this code, if disabled clusters and identities list is empty.

vrushalijoshi (approved):
looks good

PR #399, opened 2025-03-12
Refactor vs routing code

shriramsharma:
Refactor vs routing code

Modified the `FQDN -> []RouteDestination` map's key from SNI FQDN to regular FQDN. 
I change it to SNI FQDN when generating the VS for ingress

Also, added FQDN as a name to each route. This I'll use to build the cache in controller later


vrushalijoshi (approved):

vrushalijoshi (approved):
looks good

PR #414, opened 2025-04-22
excluding istio-system ns for incluster vs and dr

shriramsharma:
excluding istio-system ns for incluster vs and dr


Punakshi (approved):

vrushalijoshi (approved):
looks good

PR #419, opened 2025-05-05
add Trafficconfig CRD processing

psikka1:
### Checklist

### Description

[Link to related ISSUE]

Thank you!

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -109,12 +109,17 @@ func modifyServiceEntryForNewServiceOrPod(
 	defer util.LogElapsedTimeForTask(ctxLogger, "event", "", "", "", "TotalModifySETime")()
 	var modifySEerr error
 	var isServiceEntryModifyCalledForSourceCluster bool
+
+	// Admiral 2.0
 	if common.IsAdmiralOperatorMode() {
 		totalConfigWriterEvents.Increment(api.WithAttributes(
 			attribute.Key("identity").String(sourceIdentity),
 			attribute.Key("environment").String(env),
 		))
 	}
+
+	ctx = context.WithValue(ctx, common.TrafficConfigIdentity, sourceIdentity)
+
 	// Assigns sourceIdentity, which could have the partition prefix or might not, to the partitionedIdentity
 	// Then, gets the non-partitioned identity and assigns it to sourceIdentity. sourceIdentity will always have the original/non-partitioned identity
 	partitionedIdentity := sourceIdentity
@@ -135,10 +140,13 @@ func modifyServiceEntryForNewServiceOrPod(
 		return nil, fmt.Errorf(common.CtxLogFormat, event, env, sourceIdentity, "", "processing skipped during cache warm up state for env="+env+" identity="+sourceIdentity)
 	}
 
+	// Admiral 2.0
 	registryConfig := registry.IdentityConfig{
 		IdentityName: sourceIdentity,
 		Clusters:     make(map[string]*registry.IdentityConfigCluster),
 	}
+	// End admiral 2.0
+
 	ctxLogger.Infof(common.CtxLogFormat, event, "", "", "", "processing")
 	var (
 		cname                                 string
@@ -192,18 +200,20 @@ func modifyServiceEntryForNewServiceOrPod(
 		return nil, nil
 	}
 
+	// This is done for optimizing dependency handler. If a new asset onboards to a service, we want to create/update the SE only in that dependent's cluster.
+	// START dependency handler optimization
 	var createResourcesOnlyInDependentOverrideClusters bool
 	dependentClusterOverride, ok := ctx.Value(common.DependentClusterOverride).(*common.Map)
 	if !ok {
 		ctxLogger.Warnf(common.CtxLogFormat, "event", "", "", "", "dependent cluster override not passed")
 	} else {
-		if dependentClusterOverride != nil && len(dependentClusterOverride.GetKeys()) > 0 {
+		if dependentClusterOverride != nil && len(dependentClusterOverride.GetValues()) > 0 {
 			ctxLogger.Infof(common.CtxLogFormat, "modifyServiceEntryForNewServiceOrPod", "", "", "", "dependent cluster override passed")
 			createResourcesOnlyInDependentOverrideClusters = true
 		}
 	}
+	// END dependency handler optimization
 
-	// build service entry spec
 	for _, clusterId := range clusters {
 		rc := remoteRegistry.GetRemoteController(clusterId)
 		if rc == nil {
@@ -224,6 +234,7 @@ func modifyServiceEntryForNewServiceOrPod(
 			continue
 		}
 
+		// START - Admiral 2.0
 		ingressEndpoint, port, _ := getOverwrittenLoadBalancer(ctxLogger, rc, clusterName, remoteRegistry.AdmiralCache)
 
 		registryConfig.Clusters[clusterId] = &registry.IdentityConfigCluster{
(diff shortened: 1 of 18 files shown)


admiral/pkg/clusters/serviceentry.go
```diff
@@ -192,18 +200,20 @@ func modifyServiceEntryForNewServiceOrPod(
...
 
+	// This is done for optimizing dependency handler. If a new asset onboards to a service, we want to create/update the SE only in that dependent's cluster.
+	// START dependency handler optimization
 	var createResourcesOnlyInDependentOverrideClusters bool
 	dependentClusterOverride, ok := ctx.Value(common.DependentClusterOverride).(*common.Map)
 	if !ok {
 		ctxLogger.Warnf(common.CtxLogFormat, "event", "", "", "", "dependent cluster override not passed")
 	} else {
-		if dependentClusterOverride != nil && len(dependentClusterOverride.GetKeys()) > 0 {
+		if dependentClusterOverride != nil && len(dependentClusterOverride.GetValues()) > 0 {
```
shriramsharma, line 210: why this change made?

admiral/pkg/clusters/trafficconfig_handler.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Deleted method to handle deletions of TrafficConfig
+func (th *TrafficConfigHandler) Deleted(ctx context.Context, obj *v1.TrafficConfig) error {
+	log.Debugf(LogFormat, common.Delete, common.TrafficConfigResourceType, obj.Name, "", common.ReceivedStatus)
+	return th.HandleTrafficConfigRecord(ctx, obj, th.RemoteRegistry, admiral.Delete)
+}
+
+// findSourceClustersForIdentity returns a list of clusters where the given identity and environment exists
+func findSourceClustersForIdentity(remoteRegistry *RemoteRegistry, identity string) []string {
```
shriramsharma, line 49: maybe I'm missing something. Do we need to iterate through all the clusters?
We can get the cluster from the `IdentityClusterCache` right?

admiral/pkg/clusters/serviceentry.go
psikka1: Its because [this method](https://github.com/istio-ecosystem/admiral/blob/f7338904d9325d0576fde641cae1e72966191fef/admiral/pkg/controller/common/types.go#L422-L430) returns the values instead of the keys. I think this was a bug but it continued to work as the Cache entries  have the same key and value. In my case the cache entry looks like this - 

```
asset -> env -> workloadEnv -> warmupDuration
```
When I called getKeys() I expected an array of workloadEnvs, but I was getting the list of warmupDurations.

admiral/pkg/clusters/trafficconfig_handler.go
```diff
@@ -0,0 +1,182 @@
...
+				// set the value for slowStart for this workloadEnv to blank
+				remoteRegistry.AdmiralCache.SlowStartConfigCache.Put(assetEnv, assetEnv, cachedEnv, "")
+			}
+		})
+	}
+
+	// Update the warmupDuration in the .local destinationRule
+	for _, cachedWorkloadEnv := range slowStartConfigsForAllWorkloadEnvs.GetKeys() {
+		// Find source clusters where this identity exists
+		sourceClusters := findSourceClustersForIdentity(remoteRegistry, assetAlias)
```
Punakshi, line 143: We can just get the first cluster from identityclustercache

admiral/pkg/controller/admiral/controller.go
```diff
@@ -354,6 +355,10 @@ func (c *Controller) processNextItem() bool {
 	if err == nil {
 		// No error, forget item
 		c.queue.Forget(item)
+		err := c.delegator.UpdateProcessItemStatus(item.(InformerCacheObj).obj, common.Processed)
```
Punakshi, line 358: This should be already happening in processItem function. Any particular reason to include it here again?

Punakshi:
Overall logic looks good. Just added 2 comments.

admiral/pkg/clusters/trafficconfig_handler.go
psikka1: Updated.

admiral/pkg/clusters/trafficconfig_handler.go
```diff
@@ -0,0 +1,163 @@
...
+			// If this env is not in the current TrafficConfig's selectors, add it to missing list
+			if _, exists := workloadEnvSelectorsMap[cachedEnv]; !exists {
+				workloadEnvsNotFoundInTrafficConfig = append(workloadEnvsNotFoundInTrafficConfig, cachedEnv)
+				log.Infof("Found workload environment in cache that is not in TrafficConfig: asset=%s, env=%s",
+					assetAlias, cachedEnv)
+				// set the value for slowStart for this workloadEnv to blank
+				remoteRegistry.AdmiralCache.SlowStartConfigCache.Put(assetEnv, assetEnv, cachedEnv, "")
+			}
+		})
+	}
```
shriramsharma, line 119: just a thought, is it possible to move this code which basically keeping the `SlowStartConfigCache` updated in a separate func? 
That func can then be called from `Added` `Updated` or `Deleted` method of the handler.
WDYT?

admiral/pkg/clusters/trafficconfig_handler.go
```diff
@@ -0,0 +1,163 @@
...
+		// pass in the first sourceCluster in the list. This is required as modifySE relies on
+		sourceCluster := sourceClusters[0]
+		ctx := context.WithValue(ctx, common.ClusterName, sourceCluster)
+		ctx = context.WithValue(ctx, common.TrafficConfigContextWorkloadEnvKey, cachedWorkloadEnv)
+
+		log.Infof("Applying TrafficConfig %s to identity %s in environment %s for cluster %s",
+			obj.Name, assetAlias, cachedWorkloadEnv, sourceCluster)
+
+		// Call modify service entry for each cluster where this identity exists
+		_, err := modifyServiceEntryForNewServiceOrPod(ctx, eventType, cachedWorkloadEnv, assetAlias, remoteRegistry)
```
shriramsharma, line 140: do we need to call it for each sourceCluster? I think modifySE runs a loop for each cluster where the identity exists.

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -1555,8 +1569,53 @@ func addUpdateRoutingDestinationRule(
...
+	drObj *networkingV1Alpha3.DestinationRule, drName string, sourceCluster string) error {
+	assetConfigMap := remoteRegistry.AdmiralCache.SlowStartConfigCache.Get(assetKey)
+	if assetConfigMap == nil {
+		return nil
 	}
 
+	// Use a function to allow breaking from the Range method
+	var warmupDurationValue int64
+	var err error
+	assetConfigMap.Range(func(envKey string, envConfig *common.Map) {
```
shriramsharma, line 1591: hey man, sorry I'm probably not understanding this. Looks like we are looping over assetConfig which has a map of workloadEnv and each workloadEnv has duration associated with it. But within the loop we keep overriding `warmupDurationValue` for each workloadEnv. 
Do all workloadEnv have the same warmupDuration? If yes, then we can just get one and break the loop right?

psikka1: Good catch, added a variable to exit early.

admiral/pkg/clusters/trafficconfig_handler.go
psikka1: done.

shriramsharma (approved):

Punakshi (approved):

PR #427, opened 2025-05-21
fixed dr pinning for custom vs in identity ns

shriramsharma:
fixed dr pinning for custom vs in identity ns

### DIFF
--- a/admiral/pkg/clusters/serviceentry.go
+++ b/admiral/pkg/clusters/serviceentry.go
@@ -1570,7 +1570,7 @@ func AddServiceEntriesWithDrWorker(
 
 		// The below code checks if there is custom VS in the identity's namespace
 		// Any Argo VS will be ignored as they are not added to the IdentityNamespaceVirtualServiceCache
-		if doDRUpdateForInClusterVSRouting && isServiceEntryModifyCalledForSourceCluster {
+		if doDRUpdateForInClusterVSRouting {
 			if eventNamespace != "" {
 				virtualServicesInIdentityNamespace := rc.
 					VirtualServiceController.IdentityNamespaceVirtualServiceCache.Get(eventNamespace)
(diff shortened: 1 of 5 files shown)


admiral/pkg/controller/istio/virtualservice.go
```diff
@@ -76,6 +77,87 @@ func NewHostToRouteDestinationCache() *HostToRouteDestinationCache {
...
+	// This is to skip Argo VS
+	for _, host := range hosts {
+		if strings.HasPrefix(host, "dummy") {
+			return nil
+		}
+	}
+	if i.cache[namespace] == nil {
+		i.cache[namespace] = make(map[string]*networking.VirtualService)
+	}
+	i.cache[namespace][name] = vs
```
Punakshi, line 136: We can evaluate if it is worth to check if the .nmesh cname of asset in this ns, matches the host in VS

admiral/pkg/clusters/serviceentry.go
```diff
@@ -1559,10 +1568,28 @@ func AddServiceEntriesWithDrWorker(
...
+							doDRUpdateForInClusterVSRouting, cluster, identityId, eventNamespace))
+				}
+			}
+		}
+
 		// This code has been added for custom VS to in-cluster VS migration
 		// We are preventing pinning to remote cluster until the custom VS's
 		// exportTo is set to dot.
-		if doDRUpdateForInClusterVSRouting {
+		if doDRUpdateForInClusterVSRouting && isServiceEntryModifyCalledForSourceCluster {
```
Punakshi, line 1592: Hey, can you clarify why wasn't this check required before?

admiral/pkg/controller/istio/virtualservice.go
Punakshi: I think vs should be at identity env lvl to avoid behavior flip across ns for same asset/env- lets take it up separately

Punakshi (approved):

PR #429, opened 2025-05-28
added active passive behavior for in cluster vs routing

shriramsharma:
added active passive behavior for in cluster vs routing

### DIFF
--- a/admiral/pkg/clusters/virtualservice_routing.go
+++ b/admiral/pkg/clusters/virtualservice_routing.go
@@ -1482,11 +1482,11 @@ func doActivePassiveInClusterVS(remoteRegistry *RemoteRegistry,
 		return nil, fmt.Errorf(
 			"the DR %s is pointing to the active cluster %s already", drName, sourceClusterLocality)
 	}
-	passiveLocality := ""
+	activeLocality := ""
 	for currentLocalityOnDR := range distribution[0].To {
-		passiveLocality = currentLocalityOnDR
+		activeLocality = currentLocalityOnDR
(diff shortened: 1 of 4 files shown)


admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -1369,6 +1381,94 @@ func processGTPAndAddWeightsByCluster(ctxLogger *log.Entry,
...
+				{
+					DnsPrefix: common.Default,
+					LbType:    model.TrafficPolicy_FAILOVER,
+					Target: []*model.TrafficGroup{
+						{
+							Region: sourceClusterLocality,
+							Weight: int32(0),
+						},
+						{
+							Region: passiveLocality,
```
Punakshi, line 1461: Shouldnt this be active locality?

admiral/pkg/clusters/virtualservice_routing.go
```diff
@@ -1755,6 +1884,23 @@ func addUpdateRoutingDestinationRule(
 			}
 		}
 
+		// We are going to add client connection pool overrides and default outlierDetection
+		// only for in-cluster DRs
+		if drNameSuffix == common.InclusterDRSuffix {
```
Punakshi, line 1889: Do we really need this check, can't we keep this configuration consistent across all DRs?

vrushalijoshi (approved):
overall looks good, just one point, need to check if we should return error from certain scenarios or just logs it.

Punakshi (approved):

PR #430, opened 2025-05-28
removed DR pinning for additional endpoints

shriramsharma:
removed DR pinning for additional endpoints


vrushalijoshi:
This will also cover canary or any other endpoints not part of GTP?

vrushalijoshi (approved):
looks good

